PR #41, opened 2024-07-29
Added conditional rendering for Galasa service name

aashir21:
# Why

- Refer to https://github.com/galasa-dev/projectmanagement/issues/1931


galasa-team:
Build successful

galasa-ui/src/app/layout.tsx
```diff
@@ -7,10 +7,13 @@ import PageHeader from '@/components/PageHeader';
 import '../styles/global.scss';
 
 export default function RootLayout({ children }: { children: React.ReactNode }) {
+
+  let galasaServiceName = process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME?.trim() || "Galasa Service";
```
eamansour: I'd set this as a `const` since it shouldn't change - also see the comment below about passing this as a property into the PageHeader component to avoid duplication

One more thing - it would be good to add `NEXT_PUBLIC_GALASA_SERVICE_NAME` to the .env file so that other developers know that this env variable is being used

galasa-ui/src/components/PageHeader.tsx
```diff
@@ -6,12 +6,16 @@
...
 
 export default function PageHeader() {
+  
+  let galasaServiceName = process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME?.trim() || "Galasa Service"
+
   return (
     <Header aria-label="Galasa Ecosystem">
       <SkipToContent />
-      <HeaderName prefix="">Galasa Ecosystem</HeaderName>
+      <HeaderName prefix="">{galasaServiceName}</HeaderName>
```
eamansour: Instead of duplicating the `galasaServiceName` variable, it would be good to pass it into the PageHeader component as a property. This will simplify the unit tests as well since we won't need to mock out the env variable.

galasa-ui/src/tests/header.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import HomePage from '@/app/page';
+import PageHeader from '@/components/PageHeader';
+import { act, render, screen } from '@testing-library/react';
+
+jest.mock('next/headers', () => ({
+  cookies: jest.fn(() => ({
+    get: jest.fn().mockReturnValue('')
+  }))
+}))
```
eamansour: Don't think we need to mock the cookies API here since we're not using it in the header - this can be removed :)

galasa-ui/src/tests/header.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+afterEach(() => {
+    delete process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME
+})
+
+test('renders Galasa Service header title when env NEXT_PUBLIC_GALASA_SERVICE_NAME is null or blank string', () => {
+
+    process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME = "";  //mocking environment variable
+    render(<PageHeader />);
+  
+    const galasaServiceName = process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME?.trim() || "Galasa Service"
```
eamansour: You can hard-code the expected values in tests instead of re-using the logic that we're testing - same goes for the tests below

galasa-ui/src/tests/header.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+jest.mock('next/headers', () => ({
+  cookies: jest.fn(() => ({
+    get: jest.fn().mockReturnValue('')
+  }))
+}))
+
+
+afterEach(() => {
+    delete process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME
+})
```
eamansour: Won't need this if the service name is given as a property to the PageHeader component

galasa-ui/src/styles/Header.module.css
eamansour, line 1: Do we need this empty file?

eamansour (changes requested):
Looking good! Just a few bits that can be improved :)

galasa-ui/src/tests/header.test.tsx
aashir21: Removed a unit test as it was no longer relevant, 

```
test('renders Galasa Service header title when env NEXT_PUBLIC_GALASA_SERVICE_NAME is a string with only spaces', () => {


    render(<PageHeader galasaServiceName=''/>);

    const titleElement = screen.getByText("Galasa Service")
    expect(titleElement.textContent).toBe("Galasa Service")
    expect(titleElement).toBeInTheDocument();

});

```

Because now that we are passing the service name as a prop a blank string will default to Galasa Service

galasa-team:
Build successful

galasa-ui/src/tests/layout.test.tsx
```diff
@@ -4,11 +4,47 @@
...
+})
+
+test('renders Galasa Service title when env NEXT_PUBLIC_GALASA_SERVICE_NAME is null or blank string', () => {
+
+  process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME = "";  //mocking environment variable
+  render(<RootLayout>
+    Hello, world!
+  </RootLayout>);
+
+  const galasaServiceName = process.env.NEXT_PUBLIC_GALASA_SERVICE_NAME?.trim() || "Galasa Service"
```
eamansour: We don't need to have the env var logic in these tests - same below

eamansour (approved):
One more comment but approved assuming it'll be addressed

galasa-team:
Build successful

PR #53, opened 2024-09-11
Fix target blank vulnerability and dependency vulnerabilities

eamansour:
## Why?
When building the webui, 5 vulnerabilities are reported and an internal scanning tool has indicated that a link in the token response modal was missing the `rel="noopener noreferrer"` attribute. This PR resolves these vulnerabilities.


aashir21 (approved):
Looks good!

PR #61, opened 2024-10-02
Fix default markdown, add set-version script, add basic linting rules

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/1947

## Changes
- Copy the `public` folder into the webui's Docker image so that the default markdown content can be rendered correctly.
- Fix `useEffect` warning
- Add set-version script and bumped the framework version being used to 0.38.0
- Added basic linting rules (2-space indentation, semicolons, no irregular whitespace) to help maintain consistency with the existing codebase when adding new content

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 // Right now, the REST interface spec is always the same version as the galasa framework bundles.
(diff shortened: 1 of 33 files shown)


aashir21 (approved):
👍🏻

PR #67, opened 2024-11-01
Web-UI now displays login times on profile page

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/1975

### DIFF
--- a/galasa-ui/src/app/myprofile/page.tsx
+++ b/galasa-ui/src/app/myprofile/page.tsx
@@ -9,12 +9,24 @@ import { useEffect, useState } from "react";
 import { Loading, ToastNotification } from "@carbon/react";
 import styles from "../../styles/MyProfile.module.css";
 import PageTile from "@/components/PageTile";
+import { FrontEndClient } from "@/utils/interfaces/FrontEndClient";
 
 export default function MyProfilePage() {
 
   const [isLoading, setIsLoading] = useState(false);
   const [isError, setIsError] = useState(false);
-  const [loginId, setLoginId] = useState("");
+
+  const [userData, setUserData] = useState<{
+    loginId: string | null;
+    webLastLogin: string | null;
+    restApiLastLogin: string | null;
+  }>({
+    loginId: null,
+    webLastLogin: null,
+    restApiLastLogin: null,
+  });
+
+  const [clients, setClients] = useState<FrontEndClient[] | []>([])
 
   const handleFetchUserData = async () => {
 
@@ -25,8 +37,12 @@ export default function MyProfilePage() {
 
       if (response.ok) {
 
-        const data = await response.text();
-        setLoginId(data);
+        const data = await response.json();
+        setUserData(data.userData)
+
+        if(data.userData.clients){
+          setClients(data.userData.clients)
+        }
 
       }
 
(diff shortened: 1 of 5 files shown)


galasa-ui/src/app/myprofile/page.tsx
```diff
@@ -9,12 +9,24 @@ import { useEffect, useState } from "react";
...
 export default function MyProfilePage() {
 
   const [isLoading, setIsLoading] = useState(false);
   const [isError, setIsError] = useState(false);
-  const [loginId, setLoginId] = useState("");
+
+  const [userData, setUserData] = useState<{
+    loginId: string | null;
+    webLastLogin: string | null;
+    restApiLastLogin: string | null;
```
eamansour: Could this be an interface?

Instead of `string | null`, you can use `field?: string` to indicate that the field hasn't been initialised (i.e. is `string | undefined`).

galasa-ui/src/tests/routes/users.test.ts
```diff
@@ -7,12 +7,43 @@
...
+
+    async json() {
+      return JSON.parse(this.body);
+    }
+  }
+
+  return {
+    ...actualNextServer,
+    NextResponse: MockNextResponse,
+  };
```
eamansour, line 45: Is this needed? We already have tests that use NextResponse without needing to mock it out

galasa-ui/src/utils/interfaces/FrontEndClient.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+export interface FrontEndClient {
+    clientName: string;
+    lastLogin: string; // Use `Date` type if you want to work directly with Date objects
```
eamansour: Comment copied from somewhere? :)

It's a good suggestion to work with the Date directly though (if you can!)

galasa-ui/src/app/myprofile/page.tsx
```diff
@@ -53,22 +69,49 @@ export default function MyProfilePage() {
...
+                      client.clientName === "web-ui"
+                        ?
+                        "Last logged in to this web application (UTC):"
+                        :
+                        "Last logged in using a Galasa personal access token (UTC):"
+                    }
+                  </h4>
+
+                  {/* Extracting the date and time from response */}
+                  <h4>&nbsp; {client.lastLogin.substring(0, 10)} {client.lastLogin.substring(11, 16)}</h4>
```
eamansour: Maybe converting the lastLogin to a Date and using that here would be easier to read?

galasa-ui/src/app/myprofile/page.tsx
```diff
@@ -53,22 +69,49 @@ export default function MyProfilePage() {
...
-                  onCloseButtonClick={function noRefCheck() { setIsError(false);}}
-                  statusIconDescription="notification"
-                  caption="Failed to fetch user profile data."
-                  title="Internal Server Error"
-                />    
+        <ToastNotification
+          aria-label="closes notification"
+          kind="error"
+          onClose={function noRefCheck() { }}
+          onCloseButtonClick={function noRefCheck() { setIsError(false); }}
```
eamansour: onClose has an empty function, we could remove that?

noRefCheck isn't very descriptive, onCloseButtonClick could just be:
```typescript
() => { setIsError(false); }
```

eamansour (approved):
A few comments but approved

galasa-ui/src/utils/interfaces/FrontEndClient.ts
eamansour: Is the openapi-generated model not usable?

Looks like it generates the right thing:
```ts
export class FrontEndClient {
    /**
    * The name of the client (e.g., \"web-ui\" or \"rest-api\").
    */
    'clientName'?: string;
    /**
    * The timestamp of the user\'s last login on the respective client.
    */
    'lastLogin'?: Date;
...
```

PR #71, opened 2024-11-22
Added run detect-secret scripts to workflow

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2047

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -61,6 +61,12 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: "20.10.0"
+      
+      - name: Turn script into an executable
+        run: chmod +x detect-secrets.sh
+  
+      - name: Run the detect secrets script
+        run: ./detect-secrets.sh
 
       - name: Installing webui's dependencies using npm
         working-directory: ./galasa-ui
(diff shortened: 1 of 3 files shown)


detect-secrets.sh
```diff
@@ -0,0 +1,238 @@
...
+# Functions
+#-----------------------------------------------------------------------------------------
+
+function usage {
+    info "Syntax: detect-secrets.sh [OPTIONS]"
+    cat <<EOF
+Options are:
+-h | --help : Display this help text
+--module The name of the module to start building from
+--chain true/false/yes/no/y/n
```
jadecarino: This script doesn't take these flags so these should be removed from this help text.

.github/workflows/pr-build.yaml
```diff
@@ -76,6 +76,12 @@ jobs:
         working-directory: ./galasa-ui
         run: |
           npm run build
+      
+      - name: Turn script into an executable
+        run: chmod +x detect-secrets.sh
+
+      - name: Run the detect secrets script
+        run: ./detect-secrets.sh
```
jadecarino: This is a bit picky but I'd say these two steps should be moved to earlier in the workflow so the author of the PR doesn't have to wait a couple minutes until the Web UI has been built etc, before finding out if the code is clear of secrets (unless there is a reason you need to wait until after the build has completed).

jadecarino (changes requested):
A minor change to the script and a question about the GitHub Actions workflow but otherwise looks good to me, as someone who doesn't know a great deal about detect secrets, but I can't spot anything obvious.

I'm unable to run the Web UI's build-locally.sh due to the continued problems with my environment not letting me build the Web UI so a couple of questions - 

- Have you added a dummy secret to the code and checked that the tool picks it up?
- Is running the tool hooked into the build-locally.sh, as that script hasn't been changed?

jadecarino (approved):
Looks good!

PR #77, opened 2024-12-09
Iss2007 and Iss2085 - Recycle ecosystem1 after Web UI changes / Support forks in the workflows

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2007 and https://github.com/galasa-dev/projectmanagement/issues/2085.

- (Iss2007) Trigger the Helm workflow which redeploys ecosystem1 following changes to the WebUI. (Iss2085) Only call this if the build is from the galasa-dev organisation, so this doesn't fail forked repos builds of the WebUI.
- (Iss2085) For forks, run a job before attempting to start the build that checks for the presence of the required secrets. If they haven't been set up, the build fails and tells the forked user to set up the secrets in the repo.
- (Iss2085) Update the secret names from `GALASA_TEAM_WRITE_PACKAGES_TOKEN` to just `WRITE_PACKAGES_TOKEN` as galasa team is only specific to when the workflow is running in galasa-dev, not relevant for forks.
- Move detect-secrets to separate job from the PR build.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,11 +17,13 @@ env:
 
 jobs:
   check-secrets:
-    name: Check the required secrets are set in this organisation
+    # Run this job for forks
+    if: ${{ github.repository_owner != 'galasa-dev' }}
+    name: Check for required secrets
     runs-on: ubuntu-latest
 
     steps:
-      - name: Check if WRITE_PACKAGES_USERNAME is set
+      - name: Check if WRITE_PACKAGES_USERNAME is set in this repository
         run: |
           if [ -z "${{ secrets.WRITE_PACKAGES_USERNAME }}" ]; then
             echo "WRITE_PACKAGES_USERNAME is not set. Please configure it in the repository secrets."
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -12,12 +12,39 @@ on:
...
+    # Run this job for forks
+    if: ${{ github.repository_owner != 'galasa-dev' }}
+    name: Check for required secrets
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check if WRITE_GITHUB_PACKAGES_USERNAME is set in this repository
+        run: |
+          if [ -z "${{ secrets.WRITE_GITHUB_PACKAGES_USERNAME }}" ]; then
+            echo "WRITE_GITHUB_PACKAGES_USERNAME is not set. Please configure it in the repository secrets."
```
eamansour: Would it make sense to provide more of an explanation as to what `WRITE_GITHUB_PACKAGES_USERNAME` and `WRITE_GITHUB_PACKAGES_TOKEN` should be set to in the respective error messages for clarity?

eamansour (approved):
One small question but approved :)

.github/workflows/build.yaml
jadecarino: Yes good point thank you - I'll update these now

jadecarino: I have now added more detail to these comments explaining what the secrets need to contain.

PR #89, opened 2025-02-04
User now locks out if 403 forbidden

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2121


galasa-ui/src/components/HomeContent.tsx
```diff
@@ -25,14 +30,28 @@ export default function HomeContent({markdownContentPromise}: HomeContentProps)
     const setRenderedHtmlFromMarkdown = async () => {
       try {
         const markdownContent = await markdownContentPromise;
-        setRenderedHtmlContent(md.render(markdownContent));
+
+        if (typeof markdownContent === "string") {
```
techcobweb, line 34: wouldn't a switch statement be better here, with an error if markdown content was neither a string nor a 403 error ?

techcobweb (approved):

PR #92, opened 2025-02-10
WebUI now has a users list

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2127

Screenshot 2025-02-10 at 15 05 07


galasa-ui/src/app/users/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { createAuthenticatedApiConfiguration } from '@/utils/api';
+import React, { useEffect } from 'react';
+import * as Constants from "@/utils/constants";
+import BreadCrumb from '@/components/common/BreadCrumb';
+import PageTile from '@/components/PageTile';
+import UsersList from '@/components/users/UsersList';
+
+export const dynamic = 'force-dynamic';
+
+function UsersPage() {
```
eamansour, line 17: For consistency with the other pages, maybe we could define this with `export default function`?

galasa-ui/src/app/users/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const apiConfig = createAuthenticatedApiConfiguration();
+  const fetchAllUsersInEcosystem = async () => {
```
eamansour: Service instead of ecosystem? Or fetchAllUsersFromApiServer?

galasa-ui/src/app/users/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    let users : UserData[] = [];
+
+    const usersApiClient = new UsersAPIApi(apiConfig);
+    const usersReponse = await usersApiClient.getUserByLoginId(Constants.CLIENT_API_VERSION);
```
eamansour, line 25: Really minor - `getUserByLoginId` is a bit misleading since we're not passing in a login ID - if we're getting all users, shouldn't the openapi spec be `getUsers`?

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  role: string;
+  lastLogin: Date | string;
+  lastAccessTokenUse: Date | string;
+}
+
+function UsersList({ usersListPromise }: UsersListSectionProps) {
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [users, setUsers] = useState<any>([]);
```
eamansour: Could `users` be a type from the OpenAPI bean instead of `any`?

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  if(isError){
+    return <ErrorPage />;
+  }
+
+  return (
+    <div className={styles.userListContainer}>
+
+      <h3 className="margin-y-1">Users</h3>
```
eamansour: I don't think we need this heading since the page title already mentions Users

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+      id: user.id,
+      loginId: user.loginId,
+      role: user.synthetic?.role?.metadata?.name || "n/a",
+      lastLogin:
+        user.clients && user.clients.length > 0 && user.clients[0].lastLogin
+          ? new Date(user.clients[0].lastLogin).toLocaleDateString("en-GB") // use the British date format (DD/MM/YYYY)
+          : "n/a",
+      lastAccessTokenUse:
+        user.clients && user.clients.length > 1 && user.clients[1].lastLogin
+          ? new Date(user.clients[1].lastLogin).toLocaleDateString("en-GB")
```
eamansour: toLocaleDateString can be inefficient - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+interface UsersListSectionProps {
+  usersListPromise: Promise<UserData[]>;
+}
+
+interface IUser {
+  id: string;
+  loginId: string;
+  role: string;
+  lastLogin: Date | string;
+  lastAccessTokenUse: Date | string;
```
eamansour: Maybe just string if these properties hold the formatted dates?

galasa-ui/src/components/users/UsersList.tsx
eamansour: How can we guarantee that the client at index 0 is for the last login and 1 is the access token login? If the order changes, then the data will be presented incorrectly?

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+          getTableProps,
+          onInputChange
+        }: {
+          rows: any[];
+          headers: any[];
+          getHeaderProps: (options: any) => any;
+          getRowProps: (options: any) => any;
+          getTableProps: () => any;
+          onInputChange: () => any;
+        }) => (
```
eamansour: Can the table be simplified in any way? Or can parameters be typed here? Using `any` makes it tricky to know what we should be expecting these parameters to be.

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+                      {header.header}
+                    </TableHeader>
+                  ))}
+                </TableRow>
+              </TableHead>
+              <TableBody>
+                {rows.map((row) => (
+                  <TableRow key={row.id} {...getRowProps({ row })}>
+                    {row.cells.map((cell: any) => (
+                      <TableCell key={cell.id}>{cell.value}</TableCell>
```
eamansour: Same comment about using `any` - `cell` is `any` here too

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+            </Table>
+          </TableContainer>
+        )}
+      </DataTable>
+
+    </div>
+  );
+}
+
+export default UsersList;
```
eamansour: Missing unit tests for the UsersList component. Also, should this be renamed to UsersTable instead?

galasa-ui/src/app/users/page.tsx
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import { UserData, UsersAPIApi } from '@/generated/galasaapi';
+import { createAuthenticatedApiConfiguration } from '@/utils/api';
+import React, { useEffect } from 'react';
```
eamansour: useEffect isn't being used - can we remove unused imports? Same applies throughout this PR

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+"use client";
+import { User, UserData } from '@/generated/galasaapi';
```
eamansour: User isn't being used

galasa-ui/src/components/users/UsersList.tsx
```diff
@@ -0,0 +1,167 @@
...
+import React, { useEffect, useState } from 'react';
+import { Loading } from "@carbon/react";
+import { Table, TableHead, TableRow, TableBody, TableCell, TableHeader, DataTableCell,TableContainer, TableToolbarSearch,TableToolbarContent} from '@carbon/react';
```
eamansour: DataTableCell isn't being used

galasa-ui/src/app/users/page.tsx
aashir21: A separate PR for fixing this? Making changes to both framework and web-ui?

aashir21: > Really minor - `getUserByLoginId` is a bit misleading since we're not passing in a login ID - if we're getting all users, shouldn't the openapi spec be `getUsers`?

A separate PR for fixing this? Making changes to both framework and web-ui?

galasa-ui/src/app/users/page.tsx
aashir21: > For consistency with the other pages, maybe we could define this with `export default function`?

As per AI: "It’s generally not recommended because Next.js needs these exports to be determinable at build time. In some cases (or future versions) calling a function might prevent Next.js from optimizing your page as expected."

galasa-ui/src/tests/usersTable.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+});
+
+// Mock the ErrorPage component.
+jest.mock('@/app/error/page', () => {
+  function ErrorPageMock() {
+    return <div data-testid="error-page">Error Page</div>;
+  }
+  return ErrorPageMock;
+});
```
eamansour: I don't think we need to mock the error page - we can instead check that the error page was rendered

galasa-ui/src/tests/usersTable.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+// Mock the CSS module.
+jest.mock('@/styles/UsersList.module.css', () => ({
+  userListContainer: 'userListContainer'
+}));
```
eamansour: No need to mock CSS

galasa-ui/src/tests/usersTable.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      expect(screen.getByText('Login ID')).toBeInTheDocument();
+    });
+
+    // Check that the user data is rendered.
+    expect(screen.getByText('user1')).toBeInTheDocument();
+    expect(screen.getByText('admin')).toBeInTheDocument();
+
+    // Verify that the formatted dates are rendered.
+    const formattedLastLogin = new Intl.DateTimeFormat('en-GB').format(new Date('2020-01-01T00:00:00Z'));
+    const formattedLastAccessTokenUse = new Intl.DateTimeFormat('en-GB').format(new Date('2020-01-02T00:00:00Z'));
```
eamansour: Maybe just use the expected DD/MM/YYYY string instead of using the same formatting logic? e.g. "02/01/2020"

npm
eamansour, line 1: Accidentally checked in this file?

galasa-ui/src/tests/usersTable.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    // Due to the catch block (which sets isError to false), the table header is rendered even on error.
+    await waitFor(() => {
+      expect(screen.getByText('Login ID')).toBeInTheDocument();
+    });
+
+    // Verify that the error page is not rendered.
+    expect(screen.queryByTestId('error-page')).not.toBeInTheDocument();
+  });
+});
```
eamansour, line 123: Can we add tests to check that the sorting works as expected?

eamansour: Also tests to check that the filtering works as well?

eamansour (approved):

galasa-ui/src/app/users/page.tsx
eamansour: That doesn't make much sense. `function UsersPage` then `export default UsersPage` just means the function can be called in the module before it gets exported. For pages, it's just a stylistic choice and wont have an impact on optimization, that's why I suggested going with what we already have elsewhere...

PR #94, opened 2025-02-14
Users' roles can be updated

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2128

Screenshot 2025-02-14 at 14 29 58


galasa-ui/src/app/users/edit/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const usersApiClient = new UsersAPIApi(apiConfig);
+
+    const usersReponse = await usersApiClient.getUserByLoginId(Constants.CLIENT_API_VERSION, loginIdFromQueryParam);
+
+    if (usersReponse && usersReponse.length > 0) {
+      user = structuredClone(usersReponse[0]);
+    }
+
+    return user;
+  };
```
eamansour: We have a very similar version of this function elsewhere (I think it's also in the "My Profile" or "My Settings" page), maybe we could reduce the duplication by using the same function and importing it as needed?

galasa-ui/src/app/users/edit/updateUserRole/route.ts
```diff
@@ -0,0 +1,47 @@
+/*
```
eamansour: A server action might be better suited for updating users rather than a route - we can pass the action down into where the updates will be made

galasa-ui/src/app/users/page.tsx
```diff
@@ -10,35 +10,46 @@ import React from 'react';
...
+    let user: UserData= {};
+    const usersApiClient = new UsersAPIApi(apiConfig);
+    const usersReponse = await usersApiClient.getUserByLoginId(Constants.CLIENT_API_VERSION, "me");
+  
+    if(usersReponse && usersReponse.length > 0){
+      user = structuredClone(usersReponse[0]);
+    }
+  
+    return user;
+  };
```
eamansour: Again, very similar to getting a user by login ID in the edit user page - we can reduce the duplicate code

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+  useEffect(() => {
+    const loadUserProfile = async () => {
+      setIsError(false);
+      setIsLoading(true);
+
+      try {
+
+        const loadedProfile = await userProfilePromise;
+        setUserProfile(loadedProfile);
+        setRole({ id: loadedProfile.synthetic?.role?.metadata?.id!, name: loadedProfile.synthetic?.role?.metadata?.name!, description: loadedProfile.synthetic?.role?.metadata?.description! });
```
eamansour: Maybe we could just set the `role` to be `roleMetadata` and then we can pick out the ID, name, and description from there rather than copying the metadata out?

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+        console.error(err);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    loadUserProfile();
+  }, [userProfilePromise]);
+
+  const changeUserRole = (e: any) => {
```
eamansour: What's `e` here and what type is it supposed to be? Could be made clearer! (e.g. if it's an event, call it `event`)

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    try {
+      const response = await fetch('/users/edit/updateUserRole', {
+        method: "PUT",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(requestBody)
+      });
+
```
eamansour: Using a server action here might be better

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  if (isError) {
+    return <ErrorPage />;
+  }
+
+  return (
+    <div className={styles.roleDetails}>
+      <div className={styles.roleDetailsContainer}>
+        <h2>{userProfile.loginId}</h2>
```
eamansour, line 194: Shouldn't the login ID be displayed with the title like in the mockup (see https://github.com/galasa-dev/projectmanagement/issues/2098) ? e.g. `Edit user: myuser@myemail.com`?

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <h4>User Role</h4>
+        <p>The actions a user can or cannot on this Galasa service is controlled by their user role.</p>
```
eamansour: Should this say "The actions a user can or cannot **perform** on this Galasa service is controlled by their user role." (missing "perform")?

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <div className={styles.dropdownContainer}>
+          <Dropdown onChange={(e: any) => changeUserRole(e)} style={{ "width": "35%" }} size="lg" id="default" helperText={role.description} label={role.name} items={items} itemToString={(item: DropdownItem) => item ? item.name : ''} />
```
eamansour: Does `e` have a type?

galasa-ui/src/components/users/UsersTable.tsx
```diff
@@ -6,19 +6,24 @@
...
+import { Edit } from '@carbon/icons-react';
+import styles from "@/styles/UsersList.module.css";
+import Link from 'next/link';
 
 export const dynamic = "force-dynamic";
 
-interface UsersListSectionProps {
-  usersListPromise: Promise<UserData[]>;
+type UsersListPromiseType = Promise<UserData[]>;
+type CurrentUserPromiseType = Promise<UserData>;
```
eamansour: I don't think we need these type aliases - we could just use actual types in the interface

galasa-ui/src/tests/components/UserRoleSelection.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+import React from 'react';
+import { render, screen, fireEvent, waitFor, act } from '@testing-library/react';
+import UserRoleSection from '../../components/users/UserRoleSection';
+import { UserData } from '@/generated/galasaapi';
+
+// --- Mocks for Carbon Components and ErrorPage --- //
+jest.mock('@carbon/react', () => ({
+  ButtonSet: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  Button: (props: any) => <button {...props}>{props.children}</button>,
+  Dropdown: (props: any) => (
```
eamansour, line 17: Are we not able to run tests without mocking the carbon components out? Seems like we're not testing what actually appears since we're replacing the carbon components with vanilla HTML

galasa-ui/src/tests/components/UserRoleSelection.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+  InlineNotification: (props: any) => (
+    <div data-testid="notification">{props.title}</div>
+  ),
+}));
+
+jest.mock('@/app/error/page', () => {
+  const ErrorPage = () => <div data-testid="error-page">Error</div>;
+  ErrorPage.displayName = 'ErrorPage';
+  return ErrorPage;
+});
```
eamansour: Again, don't think we need to mock out the error page

galasa-ui/src/tests/components/UserRoleSelection.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+    const resetButton = screen.getByText('Reset') as HTMLButtonElement;
+    const saveButton = screen.getByText('Save') as HTMLButtonElement;
+    // Initially, both buttons are disabled.
+    expect(resetButton).toBeDisabled();
+    expect(saveButton).toBeDisabled();
+
+    // Change the dropdown selection to a different role.
+    // In our mock, "tester" is at index 0 and "admin" is at index 1.
+    const dropdown = screen.getByTestId('dropdown') as HTMLSelectElement;
+    fireEvent.change(dropdown, { target: { value: '1' } });
```
eamansour: Nice!

eamansour (commented):
Looking good, just a few suggestions :)

galasa-ui/src/components/users/UserRoleSection.tsx
aashir21: > Shouldn't the login ID be displayed with the title like in the mockup (see [galasa-dev/projectmanagement#2098](https://github.com/galasa-dev/projectmanagement/issues/2098)) ? e.g. `Edit user: myuser@myemail.com`?

Mike says we can keep it where it is. However, hes asked to increase the bottom margin

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const [userProfile, setUserProfile] = useState<UserData>({});
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [role, setRole] = useState<UserRoleMetadata>({});
+  const [isSaveBtnDisabled, setIsSaveBtnDisabled] = useState(true);
+  const [isResetBtnDisabled, setIsResetBtnDisabled] = useState(true);
+  const [isToastVisible, setIsToastVisible] = useState(false);
+
+  const items = [
+    {
```
techcobweb: These should be obtained from the REST API, not hard-coded into the web UI.

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+            itemToString={(item: DropdownItem) => (item ? item.name : '')}
+          />
+          <ButtonSet className={styles.buttonSet}>
+            <Button onClick={resetRole} disabled={isResetBtnDisabled} kind="secondary">Reset</Button>
+            <Button onClick={updateUserRole} disabled={isSaveBtnDisabled} kind="primary">Save</Button>
+          </ButtonSet>
+        </div>
+
+        {
+          isToastVisible && <InlineNotification inline={true} onClose={() => setIsToastVisible(false)} lowContrast={true} kind="success" title="Success" subtitle="User role was updated successfully." />
```
techcobweb, line 216: Wondering if the toast can be visible and then disappear after 5 seconds or so ?

galasa-ui/src/components/users/UserRoleSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+            style={{ width: "35%" }}
+            size="lg"
+            id="default"
+            helperText={role.description}
+            label="User Role" // this remains a static label for the field
+            items={items}
+            itemToString={(item: DropdownItem) => (item ? item.name : '')}
+          />
+          <ButtonSet className={styles.buttonSet}>
+            <Button onClick={resetRole} disabled={isResetBtnDisabled} kind="secondary">Reset</Button>
```
techcobweb, line 210: Reset functionality is setting the role back to what it was when the page was loaded rather than back to the role that was last saved successfully.

techcobweb (approved):

techcobweb:
Approved pending the resolution of my comments above.

PR #95, opened 2025-02-18
Users can be deleted from table

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2129

Screenshot 2025-02-18 at 16 27 54

### DIFF
--- a/galasa-ui/src/app/actions/deleteUserFromService.ts
+++ b/galasa-ui/src/app/actions/deleteUserFromService.ts
@@ -22,7 +22,6 @@ export const deleteUserFromService = async (userNumber: string) => {
   try {
 
     await usersApiClient.deleteUserByNumber(userNumber, Constants.CLIENT_API_VERSION);
-    return { status: 204, message: "User deleted." };
 
   } catch (error: any) {
     console.error(error);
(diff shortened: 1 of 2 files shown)


galasa-ui/src/app/actions/deleteUserFromService.ts
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error('User Number is required');
+  }
+
+  const apiConfig = createAuthenticatedApiConfiguration();
+  const usersApiClient = new UsersAPIApi(apiConfig);
+
+  try {
+
+    await usersApiClient.deleteUserByNumber(userNumber, Constants.CLIENT_API_VERSION);
+    return { status: 204, message: "User deleted." };
```
eamansour: Do we need to return HTTP status codes in server actions? If not, we can remove them and not have this return anything (acting like a void function which could throw an error)

galasa-ui/src/components/users/UsersTable.tsx
```diff
@@ -139,6 +150,22 @@ export default function UsersTable({ usersListPromise, currentUserPromise }: Use
 
   };
 
+  const deleteUser = async (userNumber: string) => {
+    try {
+      const response = await deleteUserFromService(userNumber);
+      if (response.status === 204) {
```
eamansour: Same comment about status codes in server actions - if they're not needed, this code could be simplified

eamansour (approved):

PR #103, opened 2025-04-25
Added a footer showing version and health

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2192

Screenshot 2025-04-25 at 16 49 53


galasa-ui/src/app/actions/healthActions.ts
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright contributors to the Galasa project
```
eamansour: Can we move this actions folder out of the app folder? The app folder is used by Next.js for routing, so it would be good to separate functions from rendered pages

galasa-ui/src/app/actions/healthActions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  try {
+
+    await bootstrapApiClient.getEcosystemBootstrap(CLIENT_API_VERSION);
+    return true;
+
+  } catch (error : any) {
+
+    console.error("Health check failed:", error);
+    return false;
```
eamansour: Not clear why this is returning true/false - should the function be called something like isGalasaServiceHealthy?

galasa-ui/src/app/actions/healthActions.ts
eamansour: Also, for getting data, don't use server actions (they prevent caching + increase network traffic since they're called with POST requests). Reserve them for updates/deletions (e.g. form submissions) - we can fetch the data in a server component and then pass things down to the footer component as properties.

galasa-ui/src/styles/Footer.module.css
```diff
@@ -0,0 +1,24 @@
+.footer {
```
eamansour, line 1: How does the footer look on smaller screens? Will the footer get squashed?

galasa-ui/src/styles/global.scss
```diff
@@ -12,6 +12,13 @@ body {
...
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+}
+
+main { 
+  flex: 1
 }
```
eamansour: A comment to explain what this `flex: 1` is for might help? Not clear what this actually does. Maybe use the long-hand flex-grow?

galasa-ui/src/components/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+"use client";
+
+import React, { useEffect, useState } from 'react';
+import styles from "@/styles/Footer.module.css";
+import { Theme } from '@carbon/react';
+import { getClientApiVersion, getServiceHealthStatus } from '@/app/actions/healthActions';
+
+export const dynamic = "force-dynamic";
```
eamansour: Do we need to disable pre-rendering here? If we pass the data down as properties, then we can remove this

galasa-ui/src/components/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const Footer = () => {
+
+  const [isHealthOk, setIsHealthOk] = useState(true);
+  const [apiVersion, setApiVersion] = useState("");
+
+  const checkServiceHealth = async () => {
+    const isPingSuccessful = await getServiceHealthStatus();
+    if (isPingSuccessful !== true) {
+      setIsHealthOk(false);
```
eamansour: setIsHealthOk(isPingSuccessful)?

galasa-ui/src/components/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  }, []);
+
+  return (
+    <Theme theme="g90">
+      <footer className={styles.footer} role="footer">
+        {
+          isHealthOk && <h6>Galasa Version {apiVersion}</h6>
+        }
+        <h6>Service Health {isHealthOk ? <div className={styles.healthy} /> : <div className={styles.error} />}</h6>
```
eamansour: Minor - Maybe use p with styling instead of h6 for the footer? h6 represents a subheading, could that throw off screen readers?

galasa-ui/src/tests/footer.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  
+});
+
+test('renders the footer with matching client api version', async () => {
+  (getServiceHealthStatus as jest.Mock).mockResolvedValue(true);
+  (getClientApiVersion    as jest.Mock).mockResolvedValue("0.37.0");
+
+  const {container} = render(<Footer/>);
+
+  const version = await screen.findByText(/0\.37\.0/);
```
eamansour: Minor - we should be able to use a string instead of regex for this findByText, which would make this slightly easier to read

galasa-ui/src/app/layout.tsx
```diff
@@ -4,8 +4,11 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 
+import Footer from '@/components/Footer';
 import PageHeader from '@/components/headers/PageHeader';
 import '@/styles/global.scss';
+import { getServiceHealthStatus } from './actions/healthActions';
+import { CLIENT_API_VERSION } from '@/utils/constants';
```
eamansour: Unused imports? These show up as warnings, would be good to avoid checking in.

galasa-ui/src/components/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+  };
+
+  useEffect(() => {
+
+    Promise.all([
+      checkServiceHealth(),
+      getApiVersion()
+    ]);
+
```
eamansour: Is Promise.all needed here? I think we could just call each function one after the other

galasa-ui/src/app/actions/healthActions.ts
eamansour: Instead of multiple returns, we could use an isServiceHealthy variable?

galasa-ui/src/tests/footer.test.tsx
```diff
@@ -0,0 +1,54 @@
+/*
```
eamansour: Move this file into the `tests/components` folder to keep the component tests organised?

galasa-ui/src/tests/footer.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styles from "@/styles/Footer.module.css";
+
+jest.mock('@/app/actions/healthActions', () => ({
+  getServiceHealthStatus: jest.fn(),
+  getClientApiVersion:     jest.fn(),
+}));
+
+test('renders the footer', () => {
+
+  render(<Footer/>);
```
eamansour: There are warnings appearing in the test logs when running these tests:
```
     console.error
      Warning: An update to Footer inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at Footer (/home/runner/work/webui/webui/galasa-ui/src/components/Footer.tsx:18:47)

      22 |     const isPingSuccessful = await getServiceHealthStatus();
      23 |     if (isPingSuccessful !== true) {
    > 24 |       setIsHealthOk(false);
         |       ^
      25 |     }
      26 |   };
      27 |
```

The render needs to be something like the following if the useEffect calls asynchronous code, which causes a change in any of the useState variables:
```ts
  await act(async () => {
    return render(<Footer/>);
  });
```

eamansour: Same applies to tests below

galasa-ui/src/tests/footer.test.tsx
eamansour: Also really minor but it would be nice to rename the file to Footer.test.tsx, just to match up with the component in Footer.tsx (again to keep consistency with other component tests)

galasa-ui/src/styles/Footer.module.css
aashir21: The footer is fully responsive on small-medium-large screens. The text flow is maintained in a row without elements smushing into one another

galasa-ui/src/actions/healthActions.ts
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+"use server";
```
eamansour: See my last comment https://github.com/galasa-dev/webui/pull/103#discussion_r2060557858 - this hasn't been resolved yet

galasa-ui/src/app/layout.tsx
```diff
@@ -22,6 +24,7 @@ export default function RootLayout({ children }: { children: React.ReactNode })
       <body>
         <PageHeader galasaServiceName={galasaServiceName} />
         {children}
+        <Footer serviceHealthyPromise={getServiceHealthStatus()} clientVersionPromise={getClientApiVersion()}/>
```
eamansour, line 27: This is good - just need to move away from using server actions for getting data (see https://github.com/galasa-dev/webui/pull/103#discussion_r2060557858)

galasa-ui/src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+      setApiVersion(apiVersion);
+    }
+  };
+
+  useEffect(() => {
+
+    checkServiceHealth();
+    getApiVersion();
+
+  }, []);
```
eamansour: A warning will appear here about useEffect missing the promises as dependencies, can this be addressed? Might need some investigation to make sure we're not constantly re-rendering the footer

eamansour (commented):
Nice changes - a couple comments remain

galasa-ui/src/components/Footer.tsx
```diff
@@ -37,15 +37,15 @@ const Footer = ({ serviceHealthyPromise, clientVersionPromise }: FooterProps) =>
...
 
-  }, []);
+  }, [serviceHealthyPromise, clientVersionPromise]); // included these to satisfy React Hook linting – their identity is stable so the effect still only runs once
 
   return (
     <Theme theme="g90">
       <footer className={styles.footer} role="footer">
         {
-          isHealthOk && <p>Galasa Version {apiVersion}</p>
+          isHealthOk ? <p>Galasa Version {apiVersion}</p> : <div/> // Added an empty div to maintain content flow and layout
```
eamansour: Do we need this empty div? Would it add extra empty space onto the footer if the service isn't healthy?

eamansour (approved):

PR #109, opened 2025-05-29
Add a mechanism to allow developers to connect their local webui with a remote Galasa service backend

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2248

## Changes
- Added `GALASA_DEV_TOKEN` env variable so that a personal access token can be used to authenticate with a remote Galasa service instead of redirecting to Dex
- Updated middleware to handle authenticating with a personal access token in **development mode only**
- Updated README on how to connect a local webui with a remote API server
- Added comments to the .env file to clarify each env variable and the purpose of the file
- Added unit tests and removed unused imports from existing unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-
 # Don't ever check-in node modules..
 **/node_modules/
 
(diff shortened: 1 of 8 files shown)


aashir21 (approved):
Looks good

PR #112, opened 2025-06-07
Implement feature flag system for experimental UI

Tasneemmohammed0:
## Why?
Refer to galasa-dev/projectmanagement#2225
## Changes
- `FeatureFlagContext`: A new client-side context provider that manages the feature flag state. It initializes its state from a server-provided prop and writes any changes back to a cookie.
- `ExperimentalFeaturesSection`: A new UI component added to the "My Settings" page, containing the checkboxes to toggle experimental features.
- `PageHeader`: Updated to conditionally render the "Test runs" link by consuming the new useFeatureFlags hook.
- `featureFlagCookies`: A new utility file was added to centralize cookie names.
- Add all unit tests related to the functionality
>Note
By creating a centralized `FeatureFlagProvider`, we now have a single, reliable source of truth for feature availability.


galasa-ui/src/components/ExperimentalFeaturesSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { FeatureFlagProvider, useFeatureFlags } from "@/contexts/FeatureFlagContext";
+import styles from "@/styles/MySettings.module.css";
+
+export default function ExperimentalFeaturesSection() { 
+  const { isFeatureEnabled, toggleFeatureFlag } = useFeatureFlags();
+
+  // Feature configuration for easier management and display
+  const featureConfig = [
+    {
+      key: 'testRuns' as const,
```
eamansour: Minor comment - it might be good to store the feature flag keys in a common place as constants so that we don't need to worry about misspelling keys or finding occurrences if a key needs to be changed

galasa-ui/src/contexts/FeatureFlagContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export const FeatureFlagContext = createContext<FeatureFlagContextType | undefined>(undefined);
+
+export const useFeatureFlags = (): FeatureFlagContextType => {
+  const context = useContext(FeatureFlagContext);
+  if (!context) {
+    throw new Error('useFeatureFlags must be used within a FeatureFlagProvider');
+  }
+  return context;
+};
```
eamansour: I think this code would benefit from some comments explaining the `useFeatureFlags` hook, it's purpose, and how it's used in case other developers are less familiar with custom hooks.

galasa-ui/src/contexts/FeatureFlagContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (initialFlags) {
+      try {
+        const parsedFlags = JSON.parse(initialFlags);
+        return { ...DEFAULT_FEATURE_FLAGS, ...parsedFlags };
+      } catch (error) {
+        console.error('Error parsing initial feature flags:', error);
+        return DEFAULT_FEATURE_FLAGS;
+      }
+    }
+    return DEFAULT_FEATURE_FLAGS;
```
eamansour: Instead of having three return statements, we could define a variable, update it where necessary, and then have one return statement at the end of this function to return the variable?

eamansour (approved):
A couple of minor comments, but overall looks great! Approved assuming the comments will be addressed.

Tasneemmohammed0:
> A couple of minor comments, but overall looks great! Approved assuming the comments will be addressed.

Tasneemmohammed0:
@eamansour Thank you for the detailed review!

PR #114, opened 2025-06-10
Initialised test run details page

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2099

Screenshot 2025-06-10 at 11 21 56
Screenshot 2025-06-10 at 11 22 03
Screenshot 2025-06-10 at 11 22 15
Screenshot 2025-06-10 at 11 22 24


galasa-ui/src/app/testruns/[slug]/page.tsx
```diff
@@ -0,0 +1,64 @@
+/*
```
Comment (eamansour) [inline galasa-ui/src/app/testruns/[slug]/page.tsx]: This page renders in `/testruns/<runid>` but the existing test runs page is in `/test-runs`, it would be worth changing to `test-runs` to avoid creating another segment for test runs.

galasa-ui/src/app/testruns/[slug]/page.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// Define an interface for the component's props
+interface TestRunProps {
+  params: {
+    slug: string;
+  };
+}
+
+// Type the props directly on the function's parameter
+const TestRun = ({ params: { slug } }: TestRunProps) => {
```
Comment (eamansour) [inline galasa-ui/src/app/testruns/[slug]/page.tsx]: For clarity, this could be renamed to `TestRunsPage`? And to maintain consistency with other pages, we could define it with `export default function TestRunsPage()`?

galasa-ui/src/app/testruns/[slug]/page.tsx
```diff
@@ -0,0 +1,64 @@
...
+  
+    if (rasArtifactResponse) {
+      runArtifacts = structuredClone(Array.from(rasArtifactResponse)); //Convert the set into array
+    }
+  
+    return runArtifacts;
+  };
+
+  return (
+    <TestRunDetails runId={slug} runDetailsPromise={fetchRunDetailsFromApiServer()} runArtifactsPromise={fetchTestArtifacts()} runLogPromise={fetchRunDetailLogs()}/>
```
Comment (eamansour) [inline galasa-ui/src/app/testruns/[slug]/page.tsx]: Wondering if it would make more sense to await the data in this component (making it an `async` function), and passing the run details down directly from the server rather than the promises?

Might make this easier to maintain rather than passing promises and handling them in client components. If this works, we can revisit other components in another PR to switch to awaiting data in server components where applicable.

galasa-ui/src/components/common/StatusCheck.tsx
```diff
@@ -0,0 +1,30 @@
...
+        status === "Passed" ? (
+          <p className={styles.status}><CheckmarkFilled size={20} color='#24A148' /> {status}</p>
+        ) : (status === "Failed" || status === "Hung" || status === "EnvFail") ? (
+          <p className={styles.status}><ErrorOutline size={20} color='#da1e28' /> {status}</p>
+        ) : (status === "Cancelled" || status === "Ignored") ? (
+          <p className={styles.status}><ErrorOutline size={20} color='#6f6f6f' /> {status}</p>
+        ) : (status === "Requeued") ? (
+          <p className={styles.status}><Renew size={20} color='#0043ce' /> {status}</p>
+        ) : (
+          <p className={styles.status}><Warning size={20} color='#f1c21b' />Unknown</p>
```
eamansour: Can these statuses be placed into a constant that we can pull from rather than hard-coding these values?

Same goes for the `color` values - these could be in a map like:
```ts
COLORS: {
  RED: "red-code-here",
}
```

Then we could refer to them like `COLORS.RED` where needed.

galasa-ui/src/components/common/TestRunBreadCrumb.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default function EditUserBreadCrumb() {
+  return (
+    <Theme theme="g10">
+      <Breadcrumb className={styles.crumbContainer}>
+        <BreadcrumbItem isCurrentPage={false} href="/">Home</BreadcrumbItem>
+        <BreadcrumbItem isCurrentPage={false} href="#">Test Runs</BreadcrumbItem>
+      </Breadcrumb>
+    </Theme>
+  );
```
eamansour, line 22: Instead of creating a new breadcrumb component for each page, could we not have one breadcrumb component and then parameterise the path? Then, we could pass in the breadcrumb path as a list into the component?

galasa-ui/src/components/runs/InlineText.tsx
```diff
@@ -0,0 +1,18 @@
...
+import styles from "@/styles/OverviewTab.module.css";
+
+const InlineText = ({ title, value }: { title: string, value: string }) => {
+  return (
+    <div className={styles.info}>
+      <h5>{title}</h5>
+      <p className={styles.infoValue}>{value}</p>
+    </div>
+  );
+};
```
eamansour: Does InlineText mean this places text on the same line? If so, would this be better suited to be a `p` tag and a `span`?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -0,0 +1,163 @@
...
+export default function LogTab({ logs }: { logs: string }) {
+
+  const [logContent, setLogContent] = useState<string>('');
+  const [filteredContent, setFilteredContent] = useState<string>('');
+  const [searchTerm, setSearchTerm] = useState<string>('');
+  const [filters, setFilters] = useState({
+    ERROR: false,
+    WARN: false,
+    DEBUG: false,
+    INFO: false,
```
eamansour, line 23: TRACE level logging is missing here?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    return lines.map((line, index) => {
+      // Parse log line to determine level
+      let level = 'INFO';
+      if (line.includes('ERROR')) {
+        level = 'ERROR';
+      } else if (line.includes('WARN')) {
+        level = 'WARN';
+      } else if (line.includes('DEBUG')) {
+        level = 'DEBUG';
```
eamansour, line 51: Minor - we could use a map constant to refer to the different log levels?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -0,0 +1,163 @@
...
+      // Apply appropriate class based on log level
+      const levelClass = level.toLowerCase();
+      const colorClass = styles[levelClass as keyof typeof styles] || ""; //select color class based on line type (e.g if error then make the color red for that line)
+
+      return (
+        <div
+          key={index}
+          className={`${colorClass} ${styles.logEntry}`}
+        >
+          <pre>    {index + 1}.   {line}</pre>
```
eamansour, line 65: This changes the run log slightly by adding a number to the start of each line. Is there a way to keep the log lines as-is and have the line numbers appear in a separate column?

Rendering line numbers also moves the rest of the line when reaching double/triple+ digits, so formatting is affected as well.

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -0,0 +1,163 @@
...
+      });
+    }
+    // If no filters are active, show all lines (don't filter by log level)
+  
+    setFilteredContent(lines.join('\n'));
+  }, [searchTerm, logContent, filters]);
+
+  useEffect(() => {
+    setLogContent(logs);
+  }, []);
```
eamansour: This useEffect is missing a dependency on `logs`, so a warning message would appear in this file

galasa-ui/src/actions/runsAction.ts
```diff
@@ -0,0 +1,48 @@
...
+  const artifactFile = await rasApiClient.getRasRunArtifactByPath(runId, artifactUrl, CLIENT_API_VERSION);
+  const contentType = artifactFile.type;
+
+  const arrayBuffer = await artifactFile.arrayBuffer();
+  const buffer = Buffer.from(arrayBuffer);
+  const size = artifactFile.size;
+
+  const base64 = buffer.toString("base64");
+
+  let data: any;
```
eamansour: Data seems to be a string, can this be typed instead of using `any`?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  useEffect(() => {
+    if (!searchTerm.trim()) {
+      setFilteredContent(logContent);
+      return;
```
eamansour: else instead of early return?

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+        setIsError(true);
+      } finally {
+        setIsLoading(false);
+      }
+
+    };
+
+    loadRun();
+
+  }, []);
```
eamansour: I suspect this file would have some warnings due to missing useEffect dependencies

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+interface TestRunDetailsProps {
+  runId: string;
+  runDetailsPromise: Promise<Run>;
+  runLogPromise: Promise<string>;
+  runArtifactsPromise: Promise<ArtifactIndexEntry[]>;
+}
+
+// Type the props directly on the function's parameter
+const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsPromise }: TestRunDetailsProps) => {
```
eamansour: If an invalid run ID or a non-existent run ID is entered directly into the URL, what'll be rendered? I think we would want some kind of "Not Found" page to appear for the test run

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  }, []);
+
+  if (isError) {
+    return <ErrorPage />;
+  }
+
+  if (isLoading) {
+    return <Loading small={false} active={isLoading} />;
+  }
```
eamansour, line 96: Instead of a loading circle appearing over the entire screen, maybe the test run details that is still being loaded should appear as a skeleton?

That way we can render parts of the screen that can be rendered while we wait for data to appear on the page.

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+              <StatusCheck status={run?.result!}></StatusCheck>
+            </span>
+          </div>
+
+          <div>
+            <span className={styles.summaryStatus}>
+              <h5>Test: </h5>
+              <p>{run?.testName}</p>
+            </span>
+          </div>
```
eamansour: Might need to double-check but I don't think `h` tags can be used inside spans (it also doesn't make sense semantically)

Edit: Headings are block-level elements and spans are inline, which is invalid HTML.

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+        </div>
+
+        <Tabs>
+          <TabList iconSize="lg" className={styles.tabs}>
+            <Tab renderIcon={Dashboard} href="#">Overview</Tab>
+            <Tab renderIcon={Code} href="#">Methods</Tab>
+            <Tab renderIcon={CloudLogging} href="#">Run Log</Tab>
+            <Tab renderIcon={RepoArtifact} href="#">Artifacts</Tab>
```
eamansour, line 128: More of a CSS question - how does the page and the different tabs behave when you change the size of the browser window? Is it still usable if you were viewing details on a small screen or mobile device for example?

galasa-ui/src/styles/OverviewTab.module.css
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
```
eamansour: Duplicate copyright here

galasa-ui/src/tests/components/MethodsTab.test.tsx
```diff
@@ -0,0 +1,372 @@
...
+          startTime: undefined,
+          endTime: undefined,
+          status: undefined,
+          result: undefined,
+        },
+      ];
+
+      expect(() => {
+        render(<MethodsTab methods={methodsWithMissingData} />);
+      }).not.toThrow();
```
eamansour: Is this `toThrow()` method used correctly? Should it have an error attached or instead be `toThrowError()`?

galasa-ui/src/utils/functions.ts
```diff
@@ -14,4 +14,63 @@ export const handleDeleteCookieApiOperation = async (router: any) => {
     router.refresh();
 
   }
-};
\ No newline at end of file
+};
+
+export function parseIsoDateTime(isoString: string) {
+  // Construct a Date object
+  const dt = new Date(isoString);
```
eamansour, line 21: What happens if the given isoString value is not a valid datetime?

galasa-ui/src/utils/functions.ts
```diff
@@ -14,4 +14,63 @@ export const handleDeleteCookieApiOperation = async (router: any) => {
...
+ *
+ * @param startTime - the start time
+ * @param endTime - the end time
+ * @returns formatted duration string
+ */
+export function getIsoTimeDifference(startTime: string, endTime: string): string {
+  // Parse dates and get difference in seconds
+  const t1 = Date.parse(startTime);
+  const t2 = Date.parse(endTime);
+  let delta = Math.abs(t2 - t1) / 1000;  // seconds
```
eamansour: Similar comment here - what happens if startTime/endTime are not valid datetime strings?

Also t1 and t2 aren't very descriptive.

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+interface FolderNode {
+  name: string;
+  isFile: false;
+  children: { [key: string]: TreeNodeData };
+}
+
+type TreeNodeData = FileNode | FolderNode;
+
+export type DownloadResult =
```
eamansour: Does this `DownloadResult` type need to be exported?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+  function formatFileSize(bytes: number) {
+    if (bytes < 10000) {
+      setFileSize(`${bytes} bytes`);
+    } else {
+      const mb = bytes / (1024 * 1024);
+      setFileSize(`${mb.toFixed(2)} MB`);
+    }
+  }
+
+  const handleDownloadClick = () => {
```
eamansour, line 75: When downloading files, it seems to be prefixing them with a `_`, might need investigating + a unit test

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+      // result.data is either:
+      //  • a JS object (if JSON), or
+      //  • a string (if text/plain or something else)
+
+      if (result.contentType.includes('application/json')) {
+        setArtifactFile(result.data);
+      } else {
+        setArtifactFile(result.data as string); // (plain string)
+      }
```
eamansour: I think it would be better to have the `result` be a string to avoid ambiguity with its return type

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+      setContentType(result.contentType);
+      formatFileSize(result.size);
+      setBase64Data(result.base64);
```
eamansour: Instead of having separate state variables for the same downloaded file, would it be better to use a single variable that you can use to pull the relevant data out from?

That way we won't run into issues with keeping all of these variables in sync to refer to the correct downloaded file.

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+      }
+      return <pre>{JSON.stringify(data, null, 2)}</pre>;
+    }
+
+    // 4) Binary (zip, images, etc.)
+    return (
+      <p>
+        This is a binary file ({contentType}), please download it to see its content.
+      </p>
+    );
```
eamansour: Lots of return statements in this renderArtifactContent function makes it difficult to follow, maybe having one return at the bottom would be more readable?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+      contentType.includes("json") ||
+      typeof artifactFile === "object"
+    ) {
+      // if it’s a string, try to parse it first
+      let data = artifactFile;
+      if (typeof artifactFile === "string") {
+        try {
+          data = JSON.parse(artifactFile);
+        } catch {
+          // leave it as the raw string if parse fails
```
eamansour: Would it be worth logging a message in the console here?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+        }
+      }
+      return <pre>{JSON.stringify(data, null, 2)}</pre>;
```
eamansour: It's not immediately clear what the `null, 2` part in the stringify call here actually does without having to look up the function.

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+              isFile: false,
+              children: {},
+            };
+            currentNode = currentNode.children[segment] as FolderNode;
+          } else {
+            // Descend into existing folder
+            currentNode = existing as FolderNode;
+          }
+        }
+      });
```
eamansour: There's a lot going on inside this `forEach`, would be good to split into smaller functions

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+  useEffect(() => {
+    // Build the root node, which holds top-level folders and files
+    const root: FolderNode = { name: '', isFile: false, children: {} };
+
+    artifacts.forEach((artifact) => {
+      // 1) Get the raw path string, default to empty if undefined
+      const rawPath = artifact.path ?? '';
+      // 2) Remove a leading "/" or "./" if present
+      const cleaned = rawPath.replace(/^\.?\//, '');
```
eamansour: Evaluating regex could be slow, could this be simplified to an if startsWith(...)?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+      // 3) Split into segments and drop any empty strings
+      let segments = cleaned.split('/').filter((seg) => seg !== '');
+
+      // 4) If the first segment is "artifact" or "artifacts", drop it so that
+      //    "artifact/framework" becomes ["framework", ...]
+      if (segments[0]?.toLowerCase() === 'artifact' || segments[0]?.toLowerCase() === 'artifacts') {
+        segments = segments.slice(1);
+      }
```
eamansour: Minor - store the first segment in a variable to avoid the double `toLowerCase` calls?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+      // If there are no segments left, skip this artifact
+      if (segments.length === 0) {
+        return;
+      }
```
eamansour: Could we replace this early return with an if-else? This could be missed when debugging in the future

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+          const existing = currentNode.children[segment];
+
+          if (!existing) {
+            // Create new folder if it doesn’t exist
+            currentNode.children[segment] = {
+              name: segment,
+              isFile: false,
+              children: {},
+            };
+            currentNode = currentNode.children[segment] as FolderNode;
```
eamansour: Is this `as FolderNode` needed? If the value is typed when being defined, we won't need this?

Same applies below

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+      );
+    } else {
+      // Folder node: render label, then recurse on children
+      return (
+        <TreeNode isExpanded={false} key={path} id={path} label={node.name} value={node.name} renderIcon={Folder}>
+          {Object.values(node.children).map((child) => {
+            const childPath = path ? `${path}/${child.name}` : child.name;
+            return renderNode(child, childPath);
+          })}
+        </TreeNode>
```
eamansour: Could this be one return?

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+  // Recursive renderer: emits a <TreeNode> for each TreeNodeData
+  const renderNode = (node: TreeNodeData, path: string) => {
+    if (node.isFile) {
+      // Leaf file node
+      return (
+        <TreeNode
+          key={path}
+          id={path}
+          renderIcon={path.endsWith(".gz") ? Zip : Document}
```
eamansour: If an artifact ends with `.zip`, then it won't be marked as a `Zip`. Same goes for any other zip extension.

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -0,0 +1,308 @@
...
+  isFile: false;
+  children: { [key: string]: TreeNodeData };
+}
+
+type TreeNodeData = FileNode | FolderNode;
+
+export type DownloadResult =
+  | { contentType: 'application/json'; data: any; size: number; base64: string; }
+  | { contentType: 'text/plain'; data: string; size: number; base64: string; }
+  | { contentType: string; data: string; size: number; base64: string; };
```
eamansour: Would it be easier to use one type rather than three different versions based on what the contentType is?

eamansour (commented):
Good progress! I've added some initial comments. I'm aware that there are components missing unit tests, so I'll re-review when those are added.

Let me know if any of the comments need clarifying!

galasa-ui/src/components/runs/TestRunDetails.tsx
aashir21: > More of a CSS question - how does the page and the different tabs behave when you change the size of the browser window? Is it still usable if you were viewing details on a small screen or mobile device for example?

Every page is readable / accessible on smaller screens except the Artifact screen which could be made responsive using a media query

galasa-ui/src/components/runs/LogTab.tsx
eamansour: Also, if I have a log message with one level but the log message mentions another level, the line will erroneously be included in the filtered output.

So for example, an INFO level message that could say something like "blah blah DEBUG" will be included if you are filtering only on the DEBUG level.

galasa-ui/src/app/test-runs/[slug]/page.tsx
```diff
@@ -34,31 +32,34 @@ const TestRun = ({ params: { slug } }: TestRunProps) => {
...
     }
-  
+    
     return runArtifacts;
   };
 
+  // Await all the data
+  const runDetails = await fetchRunDetailsFromApiServer();
+  const runArtifacts = await fetchTestArtifacts();
+  const runLogs = await fetchRunDetailLogs();
```
Comment (eamansour) [inline galasa-ui/src/app/test-runs/[slug]/page.tsx]: No error handling here? If any of these throw errors, then we'll run into issues.

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -164,94 +171,115 @@ export function ArtifactsTab({ artifacts, runId, runName }: { artifacts: Artifac
...
-          renderIcon={path.endsWith(".gz") ? Zip : Document}
-          label={node.name}
-          value={node.name}
-          onSelect={() => downloadArtifact(runId, node.url)}
-        />
-      );
+      treeNode = <TreeNode
+        key={path}
+        id={path}
+        renderIcon={path.endsWith(".gz") ? Zip : Document}
```
eamansour, line 267: Issue remains here - any zip files that aren't `.gz` won't be rendered as zips.

galasa-ui/src/components/runs/InlineText.tsx
```diff
@@ -8,10 +8,10 @@ import styles from "@/styles/OverviewTab.module.css";
 
 const InlineText = ({ title, value }: { title: string, value: string }) => {
   return (
-    <div className={styles.info}>
-      <h5>{title}</h5>
+    <span className={styles.info}>
+      <p>{title}</p>
       <p className={styles.infoValue}>{value}</p>
-    </div>
+    </span>
```
eamansour: `p` tags are block-level and shouldn't be inside inline `span` tags.

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -118,19 +108,17 @@ const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsP
...
             <span className={styles.summaryStatus}>
               <h5>Status: </h5>
-              <p style={{"textTransform" : "capitalize"}}>{run?.status}</p>
+              <p style={{ "textTransform": "capitalize" }}>{run?.status}</p>
             </span>
             <span className={styles.summaryStatus}>
               <h5>Result: </h5>
               <StatusCheck status={run?.result!}></StatusCheck>
             </span>
           </div>
```
eamansour: A bunch of `h5` inside `span` still...

galasa-ui/src/utils/functions.ts
```diff
@@ -53,24 +62,36 @@ export function parseIsoDateTime(isoString: string) {
...
-    parts.push(`${seconds} sec${seconds !== 1 ? "s" : ""}`);
+
+  //Parse both timestamps to get time in milliseconds
+  const startedAt = Date.parse(startTime);
+  const endedAt   = Date.parse(endTime);
+
+  let result: string;
+
+  // If either parse failed, produce an error message
+  if (isNaN(startedAt) || isNaN(endedAt)) {
```
eamansour: Shouldn't we be validating the `startTime` and `endTime` values before attempting to parse them into Date objects?

galasa-ui/src/utils/functions.ts
```diff
@@ -53,24 +62,36 @@ export function parseIsoDateTime(isoString: string) {
...
+    const minutes = Math.floor(delta / 60);
+    const seconds = Math.round((delta - minutes * 60) * 10) / 10;
+
+    // Skipping zero units (except we always show seconds if everything else is zero)
+    const parts: string[] = [];
+    if (hours   > 0) parts.push(`${hours} hr${hours   !== 1 ? "s" : ""}`);
+    if (minutes > 0) parts.push(`${minutes} min${minutes !== 1 ? "s" : ""}`);
+    if (seconds > 0 || parts.length === 0) {
+      parts.push(`${seconds} sec${seconds !== 1 ? "s" : ""}`);
+    }
```
eamansour: If this is formatting the datetimes, it could be a separate function?

Also there's some inconsistency in the formatting of `if` conditions here which makes it a bit confusing to read. Would be nice to have `{}` brackets around the `if` statements for consistency.

eamansour (approved):
Needs more work but as discussed with @techcobweb, approving on the understanding that the requested changes will be delivered in separate follow-on PRs as new tasks in https://github.com/galasa-dev/projectmanagement/issues/2099.

PR #116, opened 2025-06-11
Link test runs list to details pages with fetched data

### LINKED ISSUE #2224: User can navigate from the list of test runs to the test-run-specific page.

Tasneemmohammed0:
## Why?
Closes [galasa-dev/projectmanagement#2224](https://github.com/galasa-dev/projectmanagement/issues/2224)
## Changes
1. Tab-Based Layout Implementation
    - Placeholder tabs added for:
       -  Timeframe
       - Table Design
       - Search Criteria
       - Results Tab is fully functional.
2. Added a new server action to fetch test runs submitted in the last 24 hours.
3. Results Table Implementation
    - Displays fetched test run data.
    - Supports pagination and sorting.
4. Added tests for `TestRunsTable` and `TestRunsTabs` components.


galasa-ui/src/actions/getTestRuns.ts
```diff
@@ -0,0 +1,48 @@
...
+import { createAuthenticatedApiConfiguration } from "@/utils/api";
+import * as Constants from "@/utils/constants";
+
+/**
+ * Fetches test runs from the Result Archive Store (RAS) for the last 24 hours.
+ * 
+ * @param {string} loginId - The login ID of the user.
+ * @returns {Promise<Run[]>} - A promise that resolves to an array of Run objects.
+ */
+export const fetchMyTestRunsForLastDay  = async (): Promise<Run[]> => {
```
eamansour: This function is defined as a server action, which is typically more suited towards updating/mutating data (it sends a POST request to the server behind the scenes). See https://nextjs.org/docs/app/getting-started/updating-data.

Maybe we can define this function in the TestRunsPage, pass down the Run[] data into the client component, and use a suspense or [loading.tsx](https://nextjs.org/docs/app/api-reference/file-conventions/loading) to render an initial skeleton while the server fetches the data?

eamansour: The `loginId` parameter can be removed from the comment since the function takes no parameters :)

Also, the function name says `MyTestRuns` but can now be something like `fetchAllTestRunsForLastDay`

galasa-ui/src/actions/getTestRuns.ts
```diff
@@ -0,0 +1,48 @@
...
+      
+    if(response && response.runs) {
+      const plainRuns = JSON.parse(JSON.stringify(response.runs));
+      return plainRuns as Run[];
+    }
+ 
+    return [];
+  } catch (error: any) {
+    console.error("Error fetching test runs:", error);
+    return [];
```
eamansour: Very minor - Instead of three return statements, we can use a variable and return once at the end?

galasa-ui/src/actions/getTestRuns.ts
```diff
@@ -0,0 +1,48 @@
...
+      Constants.CLIENT_API_VERSION,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      fromDate,        
+    );
+      
+    if(response && response.runs) {
+      const plainRuns = JSON.parse(JSON.stringify(response.runs));
```
eamansour: Instead of `JSON.parse(JSON.stringify(..))`, `structuredClone(response.runs)` could work?

galasa-ui/src/components/common/StatusIndicator.tsx
```diff
@@ -0,0 +1,59 @@
...
+    IconComponent = CheckmarkFilled;
+    iconClassName = styles.statusPassed;
+    break;
+      
+  case 'failed':
+  case 'envfail':
+    IconComponent = ErrorFilled;
+    iconClassName = styles.statusFailed;
+    break;
+  }
```
eamansour: This is good! We have a few other statuses that can be returned as follows:
- "Requeued": For test runs that were restarted
- "Cancelled": For test runs that were cancelled
- "Hung": For test runs that were automatically killed because they were stuck making no progress

If you think there are any icons that would be relevant for the above statuses, feel free to update this!

These statuses can be retrieved from the `/api/ras/resultnames` API endpoint.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -0,0 +1,228 @@
...
+  const handleRowClick = (runId: string) => {
+    router.push(`/test-runs/${runId}`);
+  };
+
+  if(isError) {
+    return <ErrorPage />;
+  }
+
+  if (isLoading) {
+    return <Loading description="Loading test runs..." withOverlay={false}/>;
```
eamansour: Does this loading cover the entire page? It might be nicer to use a skeleton over the table while it loads

Also see previous comment about fetching the data in a server component and then using a suspense or loading.tsx to render an initial loading skeleton.

eamansour (commented):
Looking great! Added a few comments, mostly minor. Let me know if anything doesn't make sense!

galasa-ui/src/components/common/StatusIndicator.tsx
Tasneemmohammed0: Should I make an API call or just hardcode them?

eamansour: Since we're mapping specific statuses to icons, I think hard-coding them is fine in this case.

galasa-ui/src/app/test-runs/page.tsx
```diff
@@ -3,22 +3,27 @@
...
     <main id="content">
       <BreadCrumb />
       <PageTile title={"Test Runs"} />
       <div className={styles.testRunsContentWrapper}>
-        <p className={styles.underConstruction}>
-            This page is under construction. Please come back later to query a list of test runs.
-        </p>
+        <Suspense fallback={<p>Loading...</p>}>
+          <TestRunsTabs runs={runs}/>
+        </Suspense>
```
eamansour: Have we got a unit test that makes sure the loading state appears while the page loads?

We'll need to simulate a promise that doesn't resolve while we check that `Loading...` is visible on the page.

eamansour (commented):
One more small comment, but otherwise looks good!

Happy to approve once the Galasa 0.42.0 change freeze is lifted.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -120,6 +143,14 @@ export default function TestRunsTable({rawRuns}: {rawRuns: Run[]}) {
     return text;
   }, [rawRuns]);
 
+  if (isError) {
+    return <ErrorPage />;
+  }
+
+  if (isLoading) {
+    return <Loading small={false} active={isLoading}/>;
+  }
```
eamansour: I know this is likely similar to another page, but I think showing a table skeleton while the test runs load would be nicer than a circular loading icon. See https://carbondesignsystem.com/components/data-table/usage/#loading for an example - something similar to that would be good!

Tasneemmohammed0: Nice catch! It looks cooler now.

galasa-ui/src/components/common/StatusIndicator.tsx
```diff
@@ -0,0 +1,77 @@
+/*
```
eamansour, line 1: Another PR was delivered on Friday that includes a "StatusCheck" component, which is very much similar to this "StatusIndicator" component - see https://github.com/galasa-dev/webui/blob/main/galasa-ui/src/components/common/StatusCheck.tsx

It would be good to only have one such component to avoid having multiple components that do the same thing, and I think this StatusIndicator component is easier to read and maintain.

eamansour (approved):
Approved with one more minor comment now that https://github.com/galasa-dev/webui/pull/114 was merged.

PR #121, opened 2025-06-13
Improved test run details page

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2223

In this extension:
- Created a common bread-crumb component
- Improved the search functionality (turned it into a find functionalit like IDEs)
- Redefined the run log display with columns (fixed formatting issues with row number)
- Created a 404 not found page if an invalid run id is passed.
- Handled icons for different file formats properly (JSON, JAR, ZIP etc)

Screenshot 2025-06-13 at 13 38 02
Screenshot 2025-06-13 at 13 40 32


galasa-ui/src/app/test-runs/[slug]/page.tsx
```diff
@@ -16,42 +17,59 @@ interface TestRunProps {
...
 
   const apiConfig = createAuthenticatedApiConfiguration();
 
-  let runDetails: Run | null = null;
-  let runArtifacts: ArtifactIndexEntry[] = [];
-  let runLogs: any = null;
-  let error: string | null = null;
+  const fetchRunDetailsFromApiServer = async () => {
+    try {
+      const rasApiClient = new ResultArchiveStoreAPIApi(apiConfig);
```
Comment (eamansour) [inline galasa-ui/src/app/test-runs/[slug]/page.tsx]: Are we able to make and use just one rasApiClient and then use it inside these functions, like we've done with the apiConfig?

galasa-ui/src/app/test-runs/[slug]/page.tsx
```diff
@@ -16,42 +17,59 @@ interface TestRunProps {
...
+  // Check if run exists first
+  try {
+    await fetchRunDetailsFromApiServer();
+  } catch (error: any) {
+    if (error?.code === 404) {
+      return <NotFound title='Run not found' description={`Unable to find a run with id: ${slug}`}/>;
+    } else{
+      return <ErrorPage />;
+    }
+    throw error; // Re-throw other errors
```
Comment (eamansour) [inline galasa-ui/src/app/test-runs/[slug]/page.tsx]: Where do these other errors get caught? Have we got a unit test to check the behaviour here?

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -10,12 +10,20 @@ import React from 'react';
 import { Breadcrumb, BreadcrumbItem, Theme } from "@carbon/react";
 import "@/styles/global.scss";
 import styles from "@/styles/BreadCrumb.module.css";
+import { BreadCrumbProps } from '@/utils/interfaces';
```
eamansour: Since these props are specific to the breadcrumb component, I think it would be best to keep the interface as close as possible to the component that uses it. Would be good to have the interface in this file without it being exported.

galasa-ui/src/components/common/NotFound.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
eamansour, line 6: Missing copyright

galasa-ui/src/components/common/NotFound.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export default function NotFound({ title, description }: { title: string, description: string }) {
+  return (
+    <main className='center'>
+
+      <h1>{title}</h1>
+      <p>{description}</p>
+
```
eamansour: Really minor but it might be neater to remove the extra lines around the h1 and p tags

galasa-ui/src/components/runs/ArtifactsTab.tsx
```diff
@@ -254,6 +242,23 @@ export function ArtifactsTab({ artifacts, runId, runName }: { artifacts: Artifac
...
+    const pathSplit = path.split(".");
+    const extension = pathSplit[pathSplit.length - 1]; // get the last split e.g some.file.ts -> we need the extension (ts)
+    let icon: CarbonIconType = Document;
+
+    if(extension == "gz" || extension == "zip" || extension == "jar" || extension == "rar" || extension == "7z") {
+      icon = Zip;
+    } else if (extension == "json") {
+      icon = Json;
+    } else if (extension == "png" || extension == "jpg" || extension == "jpeg" || extension == "gif" || extension == "svg") {
+      icon = Image;
```
eamansour: Checking if a file is a zip or image can be cleaned up slightly - we could have a constant list of known extensions for each (e.g. `const ZIP_EXTENSIONS = ['gz', 'zip', ...]`) and then we can do a `if (ZIP_EXTENSIONS.includes(extension))` (likewise for image extensions)

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+  const goToPreviousMatch = () => {
+    if (totalMatches > 0) {
+      setCurrentMatchIndex((prev) => (prev - 1 + totalMatches) % totalMatches);
+    }
+  };
+
+  const createSearchRegex = (term: string) => {
+    if (!term.trim()) return null;
+
+    let escapedTerm = term.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
eamansour: What does this regex do? Not immediately clear without having to dissect it!

Can this be simplified at all?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+    if (matchWholeWord) {
+      escapedTerm = `\\b${escapedTerm}\\b`;
+    }
+    const flags = matchCase ? 'g' : 'gi';
```
eamansour: Again, not immediately clear what `flags` is and what `g` and `gi` mean (assuming it relates to the regex matching settings for global matching and global case-insensitive matching).

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+  const getLogLevel = (line: string) => {
+    // Check for explicit log levels in the line
+    if (line.includes(' ERROR ')) return 'ERROR';
+    if (line.includes(' WARN ')) return 'WARN';
+    if (line.includes(' DEBUG ')) return 'DEBUG';
+    if (line.includes(' INFO ')) return 'INFO';
+    if (line.includes(' TRACE ')) return 'TRACE';
+
+    // If no explicit level found, return null (will be assigned to previous level)
+    return null;
```
eamansour: An issue remains here - if I have set a filter to only show INFO messages and I have a line in my log file that's an INFO message but the message itself contains something like `SETTING XYZ VALUE TO 'DEBUG'`, then that line will be marked as a DEBUG line (since DEBUG is above INFO in the if statements), and will not show in my filtered logs.

A unit test would probably show this happening, so we might be missing a test!

Also, one return at the bottom could be easier to read than lots of returns?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+    if (!hasActiveFilters) {
+      // none selected → hide all
+      return lines.map(line => ({ ...line, isVisible: false }));
+    }
+
+    // only show lines whose level is still checked
+    return lines.map(line => ({
+      ...line,
+      isVisible: !!filters[line.level as keyof typeof filters]
+    }));
```
eamansour: Can we use a variable here and return once at the bottom?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+  };
+
+  const goToPreviousMatch = () => {
+    if (totalMatches > 0) {
+      setCurrentMatchIndex((prev) => (prev - 1 + totalMatches) % totalMatches);
+    }
+  };
+
+  const createSearchRegex = (term: string) => {
+    if (!term.trim()) return null;
```
eamansour: Minor - could we return once at the end of this function?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+  const highlightText = (text: string, searchTerm: string, lineIndex: number) => {
+    if (!searchTerm.trim()) return text;
+
+    const regex = createSearchRegex(searchTerm);
+    if (!regex) return text;
+
+    const parts = text.split(regex);
+    const matches = text.match(regex);
+
+    if (!matches) return text;
```
eamansour: Same here regarding lots of return statements, which can make reading through this code more challenging

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+      if (parts[i]) {
+        result.push(parts[i]);
+      }
+    }
+
+    return result;
+  };
+
+  const getGlobalMatchIndex = (lineIndex: number, matchIndexInLine: number) => {
+    if (!searchTerm.trim()) return -1;
```
eamansour: Same here, can we maintain consistency and return at the end of the function?

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
+      }
+    }
+
+    return globalIndex + matchIndexInLine;
+  };
+
+  const countMatches = (lines: LogLine[], searchTerm: string) => {
+    if (!searchTerm.trim()) return 0;
+    const regex = createSearchRegex(searchTerm);
+    if (!regex) return 0;
```
eamansour, line 258: Here as well, multiple early returns.

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +54,288 @@ export default function LogTab({ logs }: { logs: string }) {
...
-      } else if (line.includes('DEBUG')) {
-        level = 'DEBUG';
-      } else if (line.includes('TRACE')) {
-        level = 'TRACE';
+    if (!processedLines.length) return null;
+
+    return processedLines.map((logLine, index) => {
+      // Skip hidden lines
+      if (!logLine.isVisible) {
+        return null;
```
eamansour: Here as well regarding early returns.

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -71,21 +72,29 @@ const TestRunDetails = ({ runId, runDetails, runLog, runArtifacts }: TestRunDeta
...
-    } finally {
-      setIsLoading(false);
-    }
+    loadUser();
 
-  }, [runDetails, runArtifacts, runLog]);
+  }, [runDetailsPromise, runArtifactsPromise, runLogPromise]);
 
   if (isError) {
     return <ErrorPage />;
```
eamansour, line 101: Minor - for the test run details page, is there a more suitable loading indicator that we can use instead of the full-page indicator that appears? Maybe like a skeleton that shows text loading?

galasa-ui/src/tests/components/LogTab.test.tsx
```diff
@@ -0,0 +1,259 @@
+import React from 'react';
```
eamansour, line 6: Missing copyright

galasa-ui/src/tests/components/LogTab.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+      render(<LogTab logs={sampleLogs} />);
+      
+      const errorCheckbox = screen.getByTestId('checkbox-error');
+      expect(errorCheckbox).toBeChecked();
+      
+      fireEvent.click(errorCheckbox);
+      expect(errorCheckbox).not.toBeChecked();
+    });
+
+    it('hides all content when all filters are unchecked', async () => {
```
eamansour, line 183: Can we have tests to check that the correct log lines are included when various filters are applied?

galasa-ui/src/tests/components/LogTab.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+describe('LogTab', () => {
+  const sampleLogs = `2024-01-01 10:00:00 INFO Starting application
+2024-01-01 10:00:01 DEBUG Initializing database connection
+2024-01-01 10:00:02 ERROR Failed to connect to database
+2024-01-01 10:00:03 WARN Connection retry attempt 1
+2024-01-01 10:00:04 INFO Application started successfully
+2024-01-01 10:00:05 TRACE Detailed execution trace
+Multi-line continuation
+2024-01-01 10:00:06 ERROR Another error occurred`;
```
eamansour, line 75: Maybe have a test that includes multiple log levels in the actual log message, then see if the line appears when filtering by the right log level?

galasa-ui/src/tests/components/LogTab.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+        const prevButton = screen.getByTestId('icon-button-previous-match');
+        
+        expect(nextButton).toBeDisabled();
+        expect(prevButton).toBeDisabled();
+      });
+    });
+  });
+
+  describe('Regular Expression Handling', () => {
+    it('escapes special regex characters in search', async () => {
```
eamansour, line 251: Can we have unit tests that try searching by regex to see if it picks up the right log lines?

galasa-ui/src/tests/routes/runs.test.ts
```diff
@@ -0,0 +1,77 @@
+/*
```
eamansour: This test file is in the routes folder? Is that right?

galasa-ui/src/tests/routes/runs.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    const element = await TestRunPage({ params: { slug } });
+
+    expect(element.type).toBe(NotFound);
+    expect(element.props).toMatchObject({
+      title: 'Run not found',
+      description: `Unable to find a run with id: ${slug}`,
+    });
+  });
+
+  it('renders <ErrorPage> when API throws code 500', async () => {
```
eamansour: What if a different non-API related error gets thrown? Add a unit test?

galasa-ui/src/utils/constants.ts
```diff
@@ -16,5 +16,40 @@ const COLORS = {
...
+      route: "/"
+    }
+
+  ],
+  EDIT_USER: [
+
+    {
+      title: "Home",
+      route: "/"
+    },
```
eamansour: To reduce the duplication of some of these breadcrumbs, we could instead format them as individual objects, then pass them as comma-separated props into components. For example:

```ts
const BREADCRUMB_ITEMS = {
    HOME = {
        title: "Home",
        route: "/"
    },
    EDIT_USER= {
        title: "Users",
        route: "/users"
    },
...
};
```

Then we could use them like:

```ts
<BreadCrumb breadCrumbItems={[HOME,EDIT_USER]}/>
```

Would that be better?

galasa-ui/src/utils/constants.ts
```diff
@@ -16,5 +16,40 @@ const COLORS = {
 
 };
 
+const BREADCRUMB_ITEMS = {
```
eamansour: Wondering if we should have separate files for related constants? Maybe organised into a constants folder? Otherwise this could become messy with lots of different constants.

galasa-ui/src/utils/functions.ts
```diff
@@ -110,3 +110,15 @@ const buildTimeDifference = (hours : number, minutes : number, seconds: number)
 
   return parts;
 };
+
+export const handleDownload = (content: string | ArrayBuffer, fileName: string) => {
```
eamansour: Similar comment here but about organising related functions into separate files, possibly under the same folder. This file already has some unrelated functions inside, like handleDeleteToken, buildTimeDifference, and now handleDownload.

galasa-ui/src/utils/interfaces.ts
```diff
@@ -67,4 +67,16 @@ export interface DataTableRow {
...
   isExpanded?: boolean;
   isSelected?: boolean;
+}
+
+export interface BreadCrumbProps {
+
+  title: string;
+  route: string;
+
+}
```
eamansour: See previous comment about moving this interface into the breadcrumb component since that's the only thing that uses it.

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -11,11 +11,25 @@ import { Breadcrumb, BreadcrumbItem, Theme } from "@carbon/react";
 import "@/styles/global.scss";
 import styles from "@/styles/BreadCrumb.module.css";
 
-function BreadCrumb() {
+export interface BreadCrumbProps {
```
eamansour: Minor - no need to export this interface if it's only used here

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -35,100 +59,331 @@ export default function LogTab({ logs }: { logs: string }) {
...
+    return escapedTerm;
+
+  };
+
+  const getLogLevel = (line: string) => {
+
+    let filteredType = null;
+
+    // This regex looks for a log level keyword ONLY if it follows a timestamp pattern at the start of the line.
+    const match = line.match(/^\s*[\d\s\-:T,.]*Z?\s?(ERROR|WARN|DEBUG|INFO|TRACE)/);
```
eamansour: Is there an alternative way of doing this without using regex? If anything goes wrong, this would be a pain to debug!

We would also need a lot of unit tests to make sure that the regex doesn't break at some point.

galasa-ui/src/components/runs/TestRunSkeleton.tsx
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import { SkeletonText, SkeletonPlaceholder, Tabs, TabList, Tab, TabPanels, TabPanel } from '@carbon/react';
+import { Dashboard, Code, CloudLogging, RepoArtifact } from '@carbon/icons-react';
+import styles from '@/styles/TestRun.module.css'; 
+import skeletonStyles from '@/styles/TestRunSkeleton.module.css';
+
+const TestRunSkeleton = () => {
```
eamansour, line 12: Instead of making this skeleton almost a duplicate of the TestRunDetails page, is there a way to tie in the loading states into the existing page?

If we add more tabs, this can become difficult to maintain, especially with the hard-coded lists like `[1, 2, 3].map` etc and the different hard-coded width values.

galasa-ui/src/styles/LogTab.module.css
```diff
@@ -15,14 +14,59 @@
...
+  position: relative;
+  display: flex;
+  align-items: center;
+  gap: 1rem;
+  flex: 1;
+}
+
+.lineNumberCol {
+  
+}
```
eamansour: lineNumberCol is empty - not needed?

galasa-ui/src/styles/TestRunSkeleton.module.css
```diff
@@ -0,0 +1,81 @@
...
+    border-radius: 4px;
+    padding: 12px;
+    background-color: #fafafa;
+  }
+  
+  .artifactPlaceholder {
+    width: 100%;
+    height: 128px;
+    margin-bottom: 8px;
+  }
```
eamansour: Minor, but it looks like these are all indented by 2 spaces unintentionally?

galasa-ui/src/components/runs/TestRunDetails.tsx
```diff
@@ -71,79 +73,83 @@ const TestRunDetails = ({ runId, runDetails, runLog, runArtifacts }: TestRunDeta
 
     setIsLoading(true);
 
-    try {
+    const loadUser = async () => {
```
eamansour: This mentions `loadUser` when it should be something about getting run details

galasa-ui/src/components/runs/LogTab.tsx
```diff
@@ -103,28 +103,39 @@ export default function LogTab({ logs }: { logs: string }) {
...
+
+    // Attempt to parse a timestamp and log level by splitting the line.
+    const tokens = line.trim().split(' ');
+    if (tokens.length >= 3 &&
+      tokens[0].length === 10 && // Simple check for DD/MM/YYYY format
+      tokens[0].charAt(2) === '/' &&
+      tokens[0].charAt(5) === '/' &&
+      tokens[1].includes(':') &&
+      tokens[1].includes('.')
+    ) {
```
eamansour, line 116: This is more readable than the original regex approach, but hard-coding specific lengths and characters could be a bit brittle if anything changes in the future (e.g. if the date format changes to `YYYY/MM/DD`, then the `charAt(2)` and `charAt(5)` will break).

Happy for you to look at improving on this in your next PR.

eamansour (approved):
Thanks for making those changes - approving as discussed to deliver these initial changes and improve upon them in the next PR.

PR #124, opened 2025-06-20
fix(i18n): Address internationalization snags and inconsistencies

shivambisen:
- Removed unnecessary Carbon icon from UI
- Loaded localized markdown file for DE locale
- Fixed layout issues caused by longer German translations
- Ensured fallback to English when feature flag is off
- Corrected pluralization and label mismatches ("Benutzeren" → "Benutzer", etc.)
- Applied proper translations in pagination controls via message catalog
- Fixed MISSING_MESSAGE error on Test Run row click
- Corrected "Access Denied" header mistakenly shown on Test Runs page

### DIFF
--- a/galasa-ui/messages/de.json
+++ b/galasa-ui/messages/de.json
@@ -111,6 +111,9 @@
     "home": "Startseite",
     "users": "Benutzer"
   },
+  "BreadCrumb": {
+    "home": "Startseite"
+  },
   "AccessDeniedModal": {
     "modalHeading": "Unzureichende Berechtigungen",
     "logoutButton": "Abmelden",
@@ -125,7 +128,6 @@
   },
   "TestRun": {
     "title": "Testläufe",
-    "underConstruction": "Diese Seite befindet sich im Aufbau. Bitte später wiederkommen.",
     "status": "Wird geladen..."
   },
   "MySettings": {
@@ -164,7 +166,7 @@
     "filter_info": "Info",
     "filter_debug": "Debug",
     "filter_trace": "Trace",
-    "download_button": "Protokoll herunterladen"
+    "download_button": "Laufprotokoll herunterladen"
   },
   "MethodsTab": {
     "title": "Methoden",
@@ -198,7 +200,7 @@
     "tabs": {
       "overview": "Übersicht",
       "methods": "Methoden",
-      "runLog": "Protokoll",
+      "runLog": "Laufprotokoll",
       "artifacts": "Artefakte"
     }
   },
@@ -209,13 +211,13 @@
     "tabs": {
       "overview": "Übersicht",
       "methods": "Methoden",
-      "runLog": "Protokoll",
+      "runLog": "Laufprotokoll",
       "artifacts": "Artefakte"
     }
   },
   "TestRunsTable": {
     "submittedAt": "Eingereicht am",
-    "testRunName": "Testlaufname",
+    "testRunName": "Name des Testlaufs",
     "requestor": "Anforderer",
     "group": "Gruppe",
     "bundle": "Bundle",
@@ -226,7 +228,7 @@
     "result": "Ergebnis",
     "submissionId":"Übermittlungs-ID",
     "noTestRunsFound": "Für den ausgewählten Zeitraum wurden keine Testläufe gefunden",
-    "loading": "Testergebnisse werden geladen...",
+    "isloading": "Testergebnisse werden geladen...",
     "limitExceededTitle": "Limit überschritten",
     "limitExceededSubtitle": "Ihre Abfrage ergab mehr als {maxRecords} Ergebnisse. Es werden die ersten {maxRecords} Datensätze angezeigt.",
     "timeFrameText": {
@@ -237,7 +239,11 @@
       "backwardText": "Vorherige Seite",
       "forwardText": "Nächste Seite",
       "itemsPerPageText": "Einträge pro Seite:",
-      "pageNumberText": "Seitennummer"
+      "pageNumberText": "Seitennummer",
+      "of": "von",
+      "items": "Einträge",
+      "pages": "Seiten",
+      "page": "Seite"
     }
   },
   "TestRunsTabs": {
--- a/galasa-ui/messages/de.json
+++ b/galasa-ui/messages/de.json
@@ -191,7 +191,7 @@
     "noTags": "Keine Tags vorhanden."
   },
   "TestRunDetails": {
-    "title": "Testlauf:",
+    "title": "Testlauf: {runName}",
     "status": "Status",
     "result": "Ergebnis",
     "test": "Test",
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -229,7 +229,7 @@
     "result": "Result",
     "submissionId": "Submission ID",
     "noTestRunsFound": "No test runs were found for the selected timeframe",
-    "loading": "Loading test results...",
+    "isloading":"Loading test results...",
     "limitExceededTitle": "Limit Exceeded",
     "limitExceededSubtitle": "Your query returned more than {maxRecords} results. Showing the first {maxRecords} records.",
     "timeFrameText": {
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -194,7 +194,7 @@
     "noTags": "No tags were associated with this test run."
   },
   "TestRunDetails": {
-    "title": "Test Run:",
+    "title": "Test Run: {runName}",
     "status": "Status",
     "result": "Result",
     "test": "Test",
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -128,7 +128,6 @@
   },
   "TestRun": {
     "title": "Test Runs",
-    "underConstruction": "This page is under construction. Please come back later to query a list of test runs.",
     "status": "Loading..."
   },
   "MySettings": {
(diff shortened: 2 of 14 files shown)


shivambisen:
missing Test Run name:

WhatsApp Image 2025-06-21 at 12 54 52 PM (1)

translations in pagination :

WhatsApp Image 2025-06-21 at 12 54 52 PM

fixed "Access Denied" header mistakenly shown on Test Runs page: 

WhatsApp Image 2025-06-21 at 12 54 51 PM (1)

delete Button layout in my setting page : 

WhatsApp Image 2025-06-21 at 12 54 51 PM

localized markdown file for DE locale:

WhatsApp Image 2025-06-21 at 12 54 46 PM

galasa-ui/messages/de.json
```diff
@@ -4,7 +4,7 @@
   },
   "Breadcrumb": {
     "home": "Startseite",
-    "users": "Benutzer",
+    "users": "Benutzeren",
```
eamansour: I know this was raised by @techcobweb, but I'm certain "Benutzeren" isn't a word. "Benutzer" can be singular and plural - e.g. "der Benutzer" (singular), "die Benutzer" (plural).

We can keep it as it was previously - same applies throughout this PR.

galasa-ui/messages/de.json
```diff
@@ -191,14 +193,14 @@
     "noTags": "Keine Tags vorhanden."
   },
   "TestRunDetails": {
-    "title": "Testlauf:",
+    "title": "Testlauf: ({runName})",
```
eamansour: Can we remove the surrounding `()` from the test run name?

galasa-ui/messages/en.json
```diff
@@ -194,7 +193,7 @@
     "noTags": "No tags were associated with this test run."
   },
   "TestRunDetails": {
-    "title": "Test Run:",
+    "title": "Test Run:({runName})",
```
eamansour: Can we remove the surrounding `()` from the test run name and add a space between "Test Run:" and the name?

galasa-ui/src/app/page.tsx
```diff
@@ -52,21 +55,28 @@ export default function HomePage() {
...
+
+    const markdownDir = path.join(process.cwd(), "public", "static", "markdown");
+
+    // Check if locale-specific file exists
+    const localizedFilePath = path.join(markdownDir, localizedFileName);
+    const fallbackFilePath = path.join(markdownDir, fallbackFileName);
+
+    const fileToRead = fs.existsSync(localizedFilePath)
+      ? localizedFilePath
+      : fallbackFilePath;
```
eamansour: This looks good - can we add a unit test to check that the correct file is selected based on the locale provided?

galasa-ui/src/app/page.tsx
```diff
@@ -75,9 +85,15 @@ export default function HomePage() {
...
-  const markdownContentPromise = fetchHomePageContentFromCps().catch(() =>
-    fetchDefaultMarkdownContent(),
-  );
+  let markdownContentPromise: Promise<MarkdownResponse>;
+
+  if(locale=="en"){
+    markdownContentPromise = fetchHomePageContentFromCps().catch(() =>
+      fetchDefaultMarkdownContent( locale ),
+    );}else{
+    markdownContentPromise = fetchDefaultMarkdownContent(locale);
```
eamansour: This surrounding `if` block isn't necessary. A user might have custom homepage content to show stored in the Galasa service's configuration properties store (CPS), so we should always try to get content from that before falling back to a default if there is no custom homepage set.

galasa-ui/src/tests/index.test.tsx
```diff
@@ -60,9 +60,10 @@ test('renders Galasa header', () => {
 
 test('renders Galasa Ecosystem homepage', async () => {
   const homePage = await act(async () => {
+    const layout = await HomePage();
```
eamansour: Very minor - but `layout` isn't quite accurate here, maybe `homePageComponent`?

eamansour (commented):
These changes are looking good! Just a few improvements that can be made. Let me know if any don't make sense!

galasa-ui/src/app/page.tsx
shivambisen: Thanks @eamansour I added the if (locale === "en") block because , CPS only contains the homepage markdown in English. So if the user selects a different locale (like "de"), we skip the CPS call and directly load the localized markdown file from the public (e.g., home-contents.de.md).

If we always call CPS first, users selecting a non-English language will still get the English content.

eamansour: Thanks for the explanation.

This behaviour is because we have a CPS property within the ecosystem1 service called `service.welcome.markdown`, which is currently set to:
```
# Welcome to your Galasa service
        To Make the most of your Galasa experience, please refer to the [Galasa documentation](https://galasa.dev/)
        ![image](https://raw.githubusercontent.com/galasa-dev/webui/ee5cf521ebb392079b3d672c78ad2da128d8fef9/galasa-ui/public/static/homeGraphic.svg)
```

This value can be customised to contain any markdown content we want. So, if some German users had their own Galasa service deployed and they set this CPS property to include content in German, then that would be rendered.

This `if` block will unfortunately break this functionality, so only users using the `en` locale would be able to customise their homepage using the CPS property value.

Hope that makes sense!

shivambisen:
@eamansour I've made the suggested changes 
Let me know if anything else needs tweaking :)

galasa-ui/messages/de.json
```diff
@@ -13,7 +13,7 @@
     "logout": "Abmelden"
   },
   "PageHeader": {
-    "users": "Benutzer",
+    "users": "Benutzeren",
```
eamansour: This "Benutzeren" still needs changing back

galasa-ui/src/utils/markdown.ts
```diff
@@ -0,0 +1,12 @@
+// src/utils/markdown.ts
```
eamansour: This file is missing Galasa's copyright statement

eamansour (commented):
Two very minor comments, but the changes look good, so I'll be happy to approve once they've been resolved and the new merge conflicts have been resolved.

shivambisen:
@eamansour Thanks for the review! I've addressed both comments and resolved the merge conflicts .

galasa-ui/messages/en.json
```diff
@@ -228,6 +227,8 @@
     "result": "Result",
     "noTestRunsFound": "No test runs were found for the selected timeframe",
     "loading": "Loading test results...",
+    "noData": "No test runs found in the last 24 hours.",
```
eamansour: We've just delivered a change that allows users to search for runs in a custom timeframe, so it no longer defaults to 24 hours.

It looks like there's a translation above called `noTestRunsFound` which we can use instead of `noData` here.

The same applies for the de messages

galasa-ui/messages/en.json
```diff
@@ -228,6 +227,8 @@
...
+    "isloading":"Loading test results...",
```
eamansour, line 231: Looks like we now have a `loading` translation and an `isloading` translation, both with the same message. It would be great if we could just have one, `isLoading` would be fine!

Again, same applies for the de messages :)

eamansour (commented):
Sorry for the hassle, just spotted two more minor issues now that the merge conflicts have been resolved! These will hopefully be my last comments :)

shivambisen:
No worries at all ,thanks for the careful review! I've addressed the issues. Please let me know if anything else needs updating

eamansour (approved):
Looks great! Thanks for making those changes - approved.

eamansour:
@shivambisen it seems like your last commit might not have been signed/verified, which is blocking us from merging this PR. Are you able to perhaps re-commit your last changes? Thanks!

shivambisen:
Hello @eamansour  Sorry for the trouble. one of my previous commits wasn’t GPG-signed. I’ve re-signed the relevant commits and force-pushed the changes. It should be good to go now. And again, sorry for the problem

eamansour:
No problem at all!

eamansour:
@shivambisen the changes have deployed on https://galasa-ecosystem1.galasa.dev/ but seem to be causing client-side exceptions, with this error:
```
 ⨯ Error: Server Functions cannot be called during initial render. This would create a fetch waterfall. Try to use a Server Component to pass data to Client Components instead.
    at PageHeaderMenu (./src/components/headers/PageHeaderMenu.tsx:44:69)
digest: "3803599244"
 GET / 500 in 109ms
```

The issue seems to be coming from this block of code in PageHeaderMenu:
```ts
  if(!isFeatureEnabled(FEATURE_FLAGS.INTERNATIONALIZATION)) {
    setUserLocale("en");
  }
```

Are you able to investigate this?


Edit: I've just opened a PR to fix this issue https://github.com/galasa-dev/webui/pull/131

shivambisen:
@eamansour  Thanks a lot for flagging this and fixing it quickly! 
Really sorry ,I completely forgot to  use useEffect for conditional rendering .
I’ll make sure to be more careful about this in the future. I believe I handled it correctly in the Theme Support PR, but must’ve missed it here. 

but i really want to learn what’s the best way to catch these kinds of issues during local development? Sometimes they don’t show up clearly when testing locally.

eamansour:
No worries @shivambisen! Issues like this would be best caught using unit tests (e.g. using jest) or integration tests (e.g. using Playwright/Cypress) - unfortunately, we don't have integration tests set up for the web UI project just yet.

Adopting test-driven development when working on new features also helps you identify issues a lot earlier. You can use the generated code coverage report to see the areas haven't been tested, which can then help you come up with tests for new scenarios (but don't write tests just to cover specific lines of code).

Of course, running the web UI locally and trying the changes out manually also helps you spot issues that you may not have picked up on in the automated tests.

Hope that helps!

shivambisen:
Got it, thanks for the tips @eamansour I’ll keep that in mind going forward! 👍

PR #125, opened 2025-06-21
Implement Interactive Time Frame Filter for Test Runs

### LINKED ISSUE #2266: Web UI: "Test Runs" page can return runs from a user-provided timeframe
## Story
As a Galasa user, I want to get a set of test runs from a given period of time on the web UI, so that I can filter out test runs that don't interest me.

## Background

Users need a way to send a custom query to the API server via the web UI to get a list of test runs from a given timeframe. See below for the initial UI mockup for this "Timeframe" tab:

Image

## Tasks
- [x] Add a "Timeframe" tab to the "Test Runs" page with controls as shown in the mockup
- [x] Set the default timeframe to get runs from the past day
  - [x] Add the `includeCursor` query parameter to `/ras/runs` requests to take advantage of performance improvements to queries
  - [x] Pass in the `cursor` value from the `/ras/runs` response into the next request to `/ras/runs` to get the next page of test runs. Do this until either the `cursor` value doesn't change or the number of test runs returned in a response is less than the number of requested test runs (i.e. `amountOfRuns` < `pageSize`) 
- [x] Unit tests for the functionality

Tasneemmohammed0:
## Why?
Closes [galasa-dev/projectmanagement#2266](https://github.com/galasa-dev/projectmanagement/issues/2266)
## Changes
**Architecture**
  - `TimeFrameContent`: The main container (brain) that manages all state and logic.
  - `TimeFrameFilter`: A layout component for the date and duration inputs.
  - `DateTimePicker`: A reusable component for selecting a date and time.
  
**Core Functionality**
  - Calculate the entire date/time model whenever the input is modified.
  - Automatically update the URL with `from` and `to` parameters on every valid state change, including initial page load.
  - Show test runs records of the specified time frame.
  - Prevent invalid date ranges and auto-correct ranges that are too long in the future.
  - Added `InlineNotification` messages to provide clear feedback to the user for errors and auto-corrections.
  - Unit and integration tests has been added to cover the new logic, including state synchronization, validation, and URL updates.
  
>[!NOTE]
>A hardcoded `maxRecords` limit of 1000 is in place to prevent front-end performance issues with large result sets. It should be addressed in a follow-up story.


galasa-ui/src/app/test-runs/page.tsx
```diff
@@ -11,53 +11,108 @@ import { HOME } from "@/utils/constants/breadcrumb";
...
+      const runsInBatch = response.runs || [];
+      if (runsInBatch.length > 0) {
+        allRuns.push(...structuredClone(runsInBatch));
+      }
+
+      // Check if the limit was exceeded
+      if (allRuns.length >= MAX_RECORDS) {
+        limitExceeded = true;
+        allRuns = allRuns.slice(0, MAX_RECORDS); // Trim to max records
+        hasMorePages = false; // Stop fetching more runs
```
eamansour: Really minor - can we change the inline comments to be above the lines they comment? That way, comments are more visible and we won't have to read the code before reading the comment.

Same applies throughout this PR where necessary

galasa-ui/src/components/test-runs/DateTimePicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  legend,
+  date,
+  time,
+  amPm,
+  onDateChange,
+  onTimeChange,
+  onAmPmChange,
+}: DateTimePickerProps) {
+  const [localTime, setLocalTime] = useState(time);
+  const invalidTimeText = 'Please enter a valid time in HH:MM format.';
```
eamansour: Now that we've added internationalization support, new text will need to be added into the files in the `messages` folder with the relevant translations.

Same applies throughout this PR wherever we've added text that can be rendered.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -192,12 +194,13 @@ export default function TestRunsTable({
...
 
-  if (!tableRows || tableRows.length === 0) {
-    return <p>No test runs found in the last 24 hours.</p>;
+  if( !tableRows || tableRows.length === 0) {
+    return <p>No test runs were found for the selected timeframe</p>;
   }
 
   return (
     <div className={styles.resultsPageContainer}>
+      {limitExceeded && <p>Your query returned more than {MAX_RECORDS} results. Showing the first {MAX_RECORDS} records.</p>}
```
eamansour: Should this be a warning message that also tells the user that they should make their query more specific to get other results?

Happy to keep as-is since we'll be working on a solution to this 1000 run limit in a different story soon.

galasa-ui/src/components/test-runs/TimeFrameContent.tsx
```diff
@@ -0,0 +1,165 @@
...
+      fromDate = combineDateTime(draftValues.fromDate, draftValues.fromTime, draftValues.fromAmPm);
+      toDate = combineDateTime(draftValues.toDate, draftValues.toTime, draftValues.toAmPm);
+    }
+
+    const { correctedFrom, correctedTo, notification: validationNotification } = applyTimeFrameRules(fromDate, toDate);
+
+    if (validationNotification?.kind === 'error') {
+      setNotification(validationNotification);
+      return; 
+    }
```
eamansour: Very minor - can we replace multiple returns with one return at the end of this function? That might make it easier to read through in the future

galasa-ui/src/components/test-runs/TimeFrameFilter.tsx
```diff
@@ -0,0 +1,51 @@
...
+        onDateChange={(date) => handleValueChange('fromDate', date)}
+        onTimeChange={(time) => handleValueChange('fromTime', time)}
+        onAmPmChange={(amPm) => handleValueChange('fromAmPm', amPm)}
+      />
+
+      <FormGroup legendText="Duration" className={styles.TimeFrameFilterItem} >
+        <div className={styles.DurationInputsContainer} key={values.toDate?.getTime() || 0}>
+          <NumberInput id="duration-days" label="Days" min={0} max={90} value={values.durationDays} onChange={(_:React.ChangeEvent<HTMLInputElement>, { value }: {value: number | string}) => handleValueChange('durationDays', value)} />
+          <NumberInput id="duration-hours" label="Hours" min={0} max={23} value={values.durationHours} onChange={(_:React.ChangeEvent<HTMLInputElement>, { value }: {value: number | string}) => handleValueChange('durationHours', value)} />
+          <NumberInput id="duration-minutes" label="Minutes" min={0} max={59} value={values.durationMinutes} onChange={(_:React.ChangeEvent<HTMLInputElement>, { value }: {value: number | string}) => handleValueChange('durationMinutes', value)} />
```
eamansour: Minor - can we put the `max` values for the durations in constants at the top of this file/inside the TimeFrameFilter component so that they're more visible?

galasa-ui/src/tests/components/TestRunsTabs.test.tsx
```diff
@@ -58,12 +68,12 @@ describe('TestRunsTabs Component', () => {
...
-    render(<TestRunsTabs runsListPromise={mockPromise}/>);
+  // test('displays the content of the Timeframe tab', () => {
+  //   render(<TestRunsTabs runsListPromise={mockPromise}/>);
 
-    fireEvent.click(screen.getByText('Timeframe'));
-    expect(screen.getByText(/Currently, all results for the last 24 hours/i)).toBeInTheDocument();
-  });
+  //   fireEvent.click(screen.getByText('Timeframe'));
+  //   expect(screen.getByText(/Currently, all results for the last 24 hours/i)).toBeInTheDocument();
+  // });
```
eamansour: Is this test needed? Does it need to be changed now that we don't only return the results in the last 24 hours?

galasa-ui/src/tests/test-runs.test.tsx
```diff
@@ -39,7 +39,7 @@ jest.mock('@/components/common/BreadCrumb', () =>
   }
 );
 
-const mockedFetch = fetchAllTestRunsForLastDay as jest.Mock;
+const mockedFetch = fetchAllTestRunsByPaging as jest.Mock;
```
eamansour, line 42: Do we have any unit tests for the `fetchAllTestRunsByPaging` function? Now that we've got some more complex logic inside, I think it would benefit from unit tests!

galasa-ui/src/utils/interfaces.ts
```diff
@@ -68,6 +68,18 @@ export interface DataTableRow {
...
+    fromDate: Date;
+    fromTime: string;
+    fromAmPm: 'AM' | 'PM';
+    toDate: Date;
+    toTime: string;
+    toAmPm: 'AM' | 'PM';
+    durationDays: number;
+    durationHours: number;
+    durationMinutes: number;
+}
```
eamansour: Minor - would it be better to organise interfaces like we do with the constants? For example, we could add an `interfaces` folder and then add separate files that group together interfaces for different parts of the webui?

Not sure if it's even a problem, just thinking that this file could get really big as more functionality is added to the webui!

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -4,6 +4,21 @@
...
+
+  const response = await fetch('/logout', { method: 'DELETE' });
+
+  if (response.status === 204) {
+
+    //auto redirect to render dex login page
+    router.refresh();
+
+  }
+};
```
eamansour: Possibly an accidental change here?

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -97,4 +112,122 @@ const buildTimeDifference = (hours : number, minutes : number, seconds: number)
...
+/**
+ *  Combines date, time, AM/PM  parts into a single, accurate Date object.
+ * 
+ * @param date - The date part as a Date object.
+ * @param time - The time part as a string in 'HH:MM' format.
+ * @param amPm - The AM/PM part as a string ('AM' or 'PM').
+ * 
+ * @return A Date object representing the combined date and time.
+ */
+export const combineDateTime = (date: Date, time: string, amPm: 'AM' | 'PM'): Date => {
```
eamansour: Minor - would it be worth making `'AM' | 'PM'` a type of its own? That way, we could easily re-use it wherever we need to

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -97,4 +112,122 @@ const buildTimeDifference = (hours : number, minutes : number, seconds: number)
...
+
+  const parts = timeString.trim().split(':');
+  if (parts.length !== 2)  return null;
+
+  const hour = parseInt(parts[0], 10);
+  const minute = parseInt(parts[1], 10);
+
+  const isValid = !isNaN(hour) && !isNaN(minute) &&
+                  hour >= 0 && hour <= 12 &&
+                  minute >= 0 && minute <= 59;
```
eamansour: There are several numbers here that could benefit from either being put into constants or explained via comments since it might not be immediately clear what they represent.

For example, `parts` is expected to be split into two (I know this means the HH and MM parts in `HH:MM`) but isn't clear what `parts` actually is. Also, `parseInt` is given `10` and it's not obvious what `10` represents without looking up the function.

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -97,4 +112,122 @@ const buildTimeDifference = (hours : number, minutes : number, seconds: number)
...
+  const amPm = hours24 >= 12 ? 'PM' : 'AM' as 'AM' | 'PM';
+
+  // Convert 24-hour format to 12-hour format for display
+  let hours12 = hours24 % 12;
+  if (hours12 === 0) {
+    hours12 = 12;
+  }
+
+  const minutesStr = minutes.toString().padStart(2, '0');
+  const hoursStr = hours12.toString().padStart(2, '0');
```
eamansour, line 149: Would it be worth explaining what the `padStart` call is doing in case other developers aren't familiar with it and to avoid them from having to look up the function?

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -97,4 +112,122 @@ const buildTimeDifference = (hours : number, minutes : number, seconds: number)
...
+ * @param {string} timeString - The string to parse.
+ * @returns {{hour: number, minute: number} | null} The parsed time parts or null if invalid.
+ */
+export const parseAndValidateTime = (timeString: string) => {
+  if(!timeString) return null;
+
+  let parsedTime = null;
+
+  const parts = timeString.trim().split(':');
+  if (parts.length !== 2)  return null;
```
eamansour: Minor - instead of another return here, we could continue if `parts.length === 2`?

eamansour (commented):
Looking great! Of course, there are other stories that will focus on sorting out issues with retrieving large amounts of test runs, but I think this is a very good start.

I've added a few comments (mostly minor ones) - let me know if any don't make sense!

galasa-ui/src/tests/test-runs.test.tsx
Tasneemmohammed0: I will include `fetchAllTestRunsByPaging` unit tests in the next pull request to ensure that all updates, including the addition of the **Search Criteria** component, are covered.

eamansour (approved):
Thanks for making those changes! Approved :)

PR #127, opened 2025-06-23
Added Theme Support -fixes #2283

shivambisen:
This PR introduces complete theme support for the Galasa Web UI using the Carbon Design System's theming capabilities.

[Closes#2283](https://github.com/galasa-dev/projectmanagement/issues/2283)

Library Used
-`carbon/react`

Features & Changes
1. Theme Selector
     -Added a ThemeSelector dropdown allowing users to switch between light and dark  themes

2. System Behavior
      -Supports fallback and overrides (e.g., forcing g90 for specific components like headers regardless of selected theme).

3. Feature Flag
        -Added THEME feature flag via FeatureFlagContext to optionally enable/disable theming support.
  
4. Testing and Fixes
      -Refactored test files to work with updated theme structure.

Demo Video -

https://github.com/user-attachments/assets/240aa075-b0d8-4113-8491-eb5816a6d092


shivambisen:
@eamansour @techcobweb  I had a quick question regarding Carbon theming. I noticed that some components require being explicitly wrapped with the `<Theme>` component to render correctly in the selected theme, while others seem to pick up the theme automatically.

eamansour:
@shivambisen This could be because we're missing a theme in the global CSS. Ideally, the theme should be applied globally via CSS, not via individual Theme components. In fact, I don't think we need to use any Theme components anywhere. A theme selector component can then cause an update to the CSS.

This tutorial goes through theme switching in carbon design quite well if you want to take a look https://developer.ibm.com/tutorials/awb-dynamic-theming-ibm-carbon/

shivambisen:
@eamansour I've completed the changes. Thanks a lot for your help! Could you please review them when you get a chance?

galasa-ui/src/app/layout.tsx
```diff
@@ -13,32 +13,36 @@ import { FeatureFlagProvider } from '@/contexts/FeatureFlagContext';
 import { cookies } from 'next/headers';
 import FeatureFlagCookies from '@/utils/featureFlagCookies';
 import { getLocale } from 'next-intl/server';
+import { ThemeProvider } from '@/contexts/ThemeContext';
 
 export const dynamic = "force-dynamic";
+type ThemeType = "white" | "g100";
```
eamansour: ThemeType is defined in `ThemeSelector`, so I think this isn't needed here.

galasa-ui/src/app/layout.tsx
```diff
@@ -13,32 +13,36 @@ import { FeatureFlagProvider } from '@/contexts/FeatureFlagContext';
...
 
-export default async function RootLayout({children,}: {children: React.ReactNode}) {
-  // Ensure that the incoming `locale` is valid
+export default async function RootLayout({ children }: { children: React.ReactNode }) {
   const locale = await getLocale();
- 
+
   const galasaServiceName = process.env.GALASA_SERVICE_NAME?.trim() || "Galasa Service";
   const featureFlagsCookie = cookies().get(FeatureFlagCookies.FEATURE_FLAGS)?.value;
+  const themeCookie = cookies().get('preferred-theme')?.value;
```
eamansour: Can we avoid using a cookie for the preferred theme?

Cookies are sent to the server on every request, which isn't ideal. I think theme-related config should be stored client-side in localStorage.

galasa-ui/src/app/layout.tsx
```diff
@@ -13,32 +13,36 @@ import { FeatureFlagProvider } from '@/contexts/FeatureFlagContext';
...
+  const initialTheme: ThemeType = themeCookie === 'g100' ? 'g100' : 'white'; // safe fallback
 
   return (
-    <html lang={locale}>
+    <html lang={locale} data-carbon-theme={initialTheme}>
```
eamansour: I don't think we need to add the theme to the `html` tag since it will only affect the body of the document?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+    startTransition(() => {
+      setTheme(selectedItem.value as ThemeType); 
+    });
+  };
+
+
+  return (
+    <div className={styles.container}>
+      <span className={styles.icon}>{translations('label')} :</span>
+      <Dropdown
```
eamansour: Instead of a dropdown, we could perhaps use a icon-based switcher with tooltips on hover to switch between dark, light, and system preference?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Dropdown } from "@carbon/react";
+import styles from "@/styles/Selector.module.css"; 
+import { useTheme } from "@/contexts/ThemeContext";
+import { useTranslations } from "next-intl";
+
+type ThemeType = "white" | "g100"; 
+
+const themes = [
+  { id: "white", text: "White", value: "white" },
+  { id: "g100", text: "Dark", value: "g100" },
```
eamansour: Instead of "White", the text should be "Light"

Also, looks like we're missing an option for setting the theme based on system preference?

galasa-ui/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+'use client';
+
+import React, { createContext, useContext, useState, useEffect } from 'react';
+
+type ThemeType = 'white' | 'g100';
```
eamansour: Light, dark, and system preference instead of white and g100?

galasa-ui/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  initialTheme?: ThemeType;
+}) => {
+  const [theme, setThemeState] = useState<ThemeType>(initialTheme);
+
+  const setTheme = (newTheme: ThemeType) => {
+    setThemeState(newTheme);
+    if (typeof window !== 'undefined') {
+      document.documentElement.setAttribute('data-carbon-theme', newTheme);
+      localStorage.setItem('preferred-theme', newTheme);
+      document.cookie = `preferred-theme=${newTheme}; path=/; max-age=${60 * 60 * 24 * 365}`;
```
eamansour: I don't think we need to set the cookie here since we're already setting the theme in localStorage

galasa-ui/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      const storedTheme = localStorage.getItem('preferred-theme') as ThemeType;
+      if (storedTheme === 'white' || storedTheme === 'g100') {
+        setTheme(storedTheme);
+      } else {
+        const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
eamansour: What if the system preference is set to prefer light mode? This will only set it to dark mode

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+];
+
+export default function ThemeSelector() {
```
eamansour, line 22: Looks like we're missing unit tests for this theme selector component

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const { theme, setTheme } = useTheme();
+  const [isPending, startTransition] = useTransition();
+  const selectedItem = themes.find((t) => t.value === theme) || themes[0];
+  const translations = useTranslations('ThemeSelector');
+  
+  const handleThemeChange = ({ selectedItem }: { selectedItem: { id: string; text: string; value: string } }) => {
+    if (!selectedItem) return;
```
eamansour: Instead of an early return here, we could invert the condition to only continue with `if (selectedItem)`?

galasa-ui/src/app/layout.tsx
```diff
@@ -13,32 +13,36 @@ import { FeatureFlagProvider } from '@/contexts/FeatureFlagContext';
...
         <title>{galasaServiceName}</title>
         <meta name="description" content="Galasa Ecosystem Web UI" />
       </head>
       <body>
         <NextIntlClientProvider>
           <FeatureFlagProvider initialFlags={featureFlagsCookie}>
-            <PageHeader galasaServiceName={galasaServiceName} />
-            {children}
-            <Footer serviceHealthyPromise={getServiceHealthStatus()} clientVersionPromise={getClientApiVersion()}/>
+            <ThemeProvider initialTheme={initialTheme}>
```
eamansour: Do we need to pass in an initial theme here? Or could the initial theme be set by this theme provider?

That way, localStorage can be used to get the theme instead of a cookie (assuming the ThemeProvider is for client-side use).

eamansour (commented):
Great work, thanks for opening this PR! Just a few suggestions that could improve things!

shivambisen:
Hi @eamansour , I’ve addressed all the review comments including theme source handling, system preference detection, and included test. Please have a look when you get a chance.

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+const themeOptions: { id: FullThemeType; label: string; icon: React.ReactNode }[] = [
+  { id: 'light', label: 'Light', icon: <Sun size={20} /> },
+  { id: 'dark', label: 'Dark', icon: <Moon size={20} /> },
+  { id: 'system', label: 'System', icon: <Laptop size={20} /> },
+];
+
+export default function ThemeSelector() {
+  const { theme, setTheme, source, setSource } = useTheme();
+  const [isPending, startTransition] = useTransition();
+  const currentTheme = source === 'system' ? 'system' : theme;
```
eamansour: Not sure if `source` is needed - are we able to treat `system` as a theme type? We're already adding it to FullThemeType, so it can be added to ThemeType, then we can perhaps remove `source`?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <Tooltip key={option.id} label={option.label} align="bottom">
+            <button
+              onClick={() => handleThemeChange(option.id)}
+              className={buttonClass}
+              disabled={isPending}
+              aria-label={option.label}
+            >
+              {option.icon}
+            </button>
+          </Tooltip>
```
eamansour: A few improvements here:

1. Instead of three separate buttons, we could have one control that, when you click on it, cycles through the different themes. The tooltip can then say something like "Switch to Dark mode" (if you're already on light mode) etc. This will save space on the menu bar. See the theme switcher on https://galasa.dev for inspiration.

2. The tooltips have a light theme regardless of the theme you're currently using, this can make it tricky to see when you're already on light mode. Would be great to have a better contrast.

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { ThemeType, useTheme } from "@/contexts/ThemeContext";
+import { Tooltip } from '@carbon/react';
+import { Sun, Moon, Laptop } from '@carbon/icons-react';
+
+type FullThemeType = ThemeType | 'system';
+
+const themeOptions: { id: FullThemeType; label: string; icon: React.ReactNode }[] = [
+  { id: 'light', label: 'Light', icon: <Sun size={20} /> },
+  { id: 'dark', label: 'Dark', icon: <Moon size={20} /> },
+  { id: 'system', label: 'System', icon: <Laptop size={20} /> },
```
eamansour: The `label` values should be added to the translatable values

eamansour (commented):
Looking better! Just a few more parts that could be improved :)

shivambisen:
@eamansour Sorry for the delay, I've made the changes and pushed the updated commits.

shivambisen:
Demo of new toggle :

https://github.com/user-attachments/assets/c983cbd4-1ff7-49e3-aac3-fae76a3ef544

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -23,9 +24,18 @@ function BreadCrumb({
...
   const translations = useTranslations("Breadcrumb");
+  const current = useTheme().theme;
+  const theme: 'g10' | 'g90' = 
+    current === 'light'
+      ? 'g10'
+      : current === 'dark'
+        ? 'g90'
+        : window.matchMedia('(prefers-color-scheme: dark)').matches
+          ? 'g90'
+          : 'g10';
```
eamansour: Minor - this is a bit tricky to read, might be better to define `theme` as a variable with `let` and then use if-else statements?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+'use client';
+
+import React, { useTransition } from "react";
+import styles from "@/styles/Selector.module.css"; 
+import { ThemeType, useTheme } from "@/contexts/ThemeContext";
+import { Sun, Moon, Laptop } from '@carbon/icons-react';
+
+type FullThemeType = ThemeType | 'system';
```
eamansour: FullThemeType isn't needed because ThemeType now also has 'system'

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const currentTheme = themeOptions[idx] || themeOptions[0];
+  const next = themeOptions[(idx + 1) % themeOptions.length];
+
+  const cycleTheme = () => {
+    startTransition(() => {
+      if (next.id === 'system') {
+        setTheme('system');
+      } else {
+        setTheme(next.id as ThemeType);
+      }
```
eamansour: This could be simplified to just `setTheme(next.id as ThemeType)` since 'system' is a ThemeType now

galasa-ui/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+const ThemeContext = createContext<ThemeContextType>({
+  theme: 'system',
+  setTheme: () => {},
+});
+
+export const ThemeProvider = ({ children }: { children: React.ReactNode }) => {
+  const [theme, setThemeState] = useState<ThemeType>('system');
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const apply = (t: ThemeType) => {
```
eamansour: Minor - `apply` can be made more descriptive, perhaps like `applyThemeToDocument`?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div className={styles.themeSwitcher}>
+      <button
+        onClick={cycleTheme}
+        className={styles.iconButton + (currentTheme.id === theme ? ` ${styles.active}` : '')}
+        disabled={isPending}
+        aria-label={currentTheme.label}
+      >
+        {currentTheme.icon}
+      </button>
```
eamansour: I think we still need to have a carbon design tooltip that appears when you hover over the button to indicate what happens when you click on the button and for accessibility purposes.

For example:
- Hovering over the light theme icon will display a tooltip that says "Switch to dark mode"
- Hovering over the dark theme icon will display a tooltip that says "Switch to system preference"
- Hovering over the system theme icon will display a tooltip that says "Switch to light mode"

eamansour (commented):
Thanks for making these changes, I think it's getting really close! I like the new button cycling - added a few more comments.

shivambisen:
Done
i have pushed the changes :)

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const cycleTheme = () => {
+    startTransition(() => {
+      setTheme(next.id as ThemeType);
+    });
+  };
+
+  return (
+    <div className={styles.themeSwitcher}>
+      <Tooltip label={next.tooltip} align="bottom">
```
eamansour: Thanks for adding this. However, the issue mentioned in https://github.com/galasa-dev/webui/pull/127#discussion_r2177956256 regarding the tooltip having a light theme regardless of the global theme remains.

Light mode (it's not obvious that this is a tooltip):
Screenshot 2025-07-04 at 11 56 07

Dark mode (looks good! Tooltip is clearly visible):
Screenshot 2025-07-04 at 11 59 33


In light mode, it would be good if the tooltip could appear more clearly by having a dark background and white text.

eamansour (commented):
Thanks for the changes, added one comment below.

One more thing: Now that the "Table Design" tab has been implemented, I can see that the dark mode doesn't render the table contents into something readable, so it's very difficult to use in dark mode.

This is what the "Table Design" tab looks like in light mode:
Screenshot 2025-07-04 at 12 07 15

And this is what it looks like in dark mode:
Screenshot 2025-07-04 at 12 03 57


Would you be able to address this issue as well? Sorry for the hassle!

Thanks in advance!

galasa-ui/src/components/headers/ThemeSelector.tsx
shivambisen: Fixed the issue where the Tooltip wasn't applying the correct theme  it wasn't reading the Carbon theme context without being wrapped in the <Theme> component.

```
<Theme theme={current} className={styles.themeContainer}>
  <Tooltip label={next.tooltip} align="bottom">
    <button
      onClick={cycleTheme}
      className={styles.iconButton + (currentTheme.id === theme ? ` ${styles.active}` : '')}
      disabled={isPending}
      aria-label={currentTheme.label}
    >
      {currentTheme.icon}
    </button>
  </Tooltip>
</Theme>
```

shivambisen:
Thanks! The first issue was caused by hardcoded background colors. I’ve removed them, and the table now renders properly in dark mode too.

eamansour:
Thanks for making those changes! I've just spotted one more area in the "Test runs" page's "Search Criteria" tab that might also be using a hard-coded text colour.

When adding a tag in the "Tags" row, the added tags appear in black text regardless of the current theme, which makes it difficult to read in dark mode:

Light mode:
Screenshot 2025-07-07 at 10 12 49

Dark mode:
Screenshot 2025-07-07 at 10 13 08


When you hover over a tag, the row turns white, which doesn't line up with the theme either:
Screenshot 2025-07-07 at 10 15 12


Looks like this is the last area that isn't rendering properly in dark mode, so it would be great if you could fix this! Feel free to flick through the web UI yourself in case you see any other areas that aren't rendering properly in dark mode. Thanks again for putting up with all of these comments :)

shivambisen:
@eamansour the changes are complete and it’s looking good now. Sorry for the delay! I also checked the other parts, and it seems there are no irregularities left. Really appreciate your attention to detail.

eamansour (approved):
Thanks for your hard work!

PR #132, opened 2025-06-27
Overview Tab has redirect link to results table

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2281


galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -10,10 +10,19 @@ import InlineText from "./InlineText";
...
 
 const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
   const tags = metadata?.tags || [];
   const translations = useTranslations("OverviewTab");
+  
+  const TWO_DAYS_GO = getTwoDaysAgo();
+  const MONTH_AGO = getOneMonthAgo();
+  const COMMON_LINK_URL = `/test-runs?runName=${metadata?.runName}&bundle=${metadata?.bundle}&group=${metadata?.group}&package=${metadata?.package}`;
+  const SAME_TEST_RUN_LINK = `${COMMON_LINK_URL}&from=${MONTH_AGO}`;
+  const OTHER_ATTEMPTS_AT_TEST = COMMON_LINK_URL + `&submissionId=${metadata?.submissionId}&from=${TWO_DAYS_GO}`;
```
eamansour: Does `OTHER_ATTEMPTS_AT_TEST` refer to getting the re-queues of the same test (e.g. RUN1-2, RUN1-3, etc.)? Or does it refer to running the same test as a different submission (e.g. a different `galasactl runs submit` command)?

If it means running the same test as a different submission, those runs may not have the same group (if no group is given in the `galasactl runs submit`, it'll get an auto-generated group ID) and will not have the same submission ID as the original test. This means that the link won't return the expected results.

Might need to clarify the name of the constant to avoid this ambiguity.

galasa-ui/src/tests/components/OverviewTab.test.tsx
```diff
@@ -6,14 +6,21 @@
...
 jest.mock('@carbon/react', () => ({
   Tag: ({ children }: { children: React.ReactNode }) => (
     <span data-testid="mock-tag">{children}</span>
   ),
+  Link: ({ children, href, renderIcon }: { 
+    children: React.ReactNode; 
+    href: string; 
+    renderIcon?: React.ComponentType 
+  }) => (
+    <a href={href} data-testid="mock-link">{children}</a>
```
eamansour, line 23: Maybe we should unit test that the right link is set? Will need to mock out the date

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -66,6 +75,15 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
             <p>{translations("noTags")}</p>
           )}
         </div>
+
+        <div className={styles.redirectLinks}>
+          <Link href={SAME_TEST_RUN_LINK} renderIcon={Launch}>
+            View other runs of this test...
+          </Link>
+          <Link href={OTHER_ATTEMPTS_AT_TEST} renderIcon={Launch}>
+            View other attempts at running this test run...
```
eamansour: The text in these links need to be internationalized in the en.json and de.json files

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -1,19 +1,46 @@
+"use client";
```
eamansour: Minor - can we move this `use client` down after the copyright statement?

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -1,19 +1,46 @@
...
+
+  const OTHER_RECENT_RUNS = `/test-runs?testName=${metadata?.testName}&bundle=${metadata?.bundle}&group=${metadata?.group}&package=${metadata?.package}&from=${MONTH_AGO}`;
+  const RETRIES_FOR_THIS_TEST_RUN = `/test-runs?submissionId=${metadata?.submissionId}&from=${weekBefore}`;
+
+  useEffect(() => {
+
+    const validateTime = () => {
+
+      const validatedTime = getAWeekBeforeSubmittedTime(metadata?.rawSubmittedAt!);
+      if (validatedTime === "Invalid date") {
```
eamansour: If the time is invalid, instead of setting an "Invalid date" value - maybe we either throw an error and catch it, or return a null value?

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -66,6 +93,21 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
         </div>
+
+        <div className={styles.redirectLinks}>
+          <Link href={OTHER_RECENT_RUNS} renderIcon={Launch}>
+            {translations("recentRunsLink")}
+          </Link>
+
+          {/* Only show the link if date is valid */}
+          {
+            weekBefore !== "Invalid date" && (
```
eamansour: See above comment about not setting a string value for an invalid value

galasa-ui/src/utils/interfaces/testRuns.ts
```diff
@@ -18,8 +18,10 @@ export interface RunMetadata {
   runName: string;
   bundle: string;
   submissionId: string;
+  package: string;
   group: string;
   requestor: string;
+  rawSubmittedAt?: string;
```
eamansour, line 24: Is this `rawSubmittedAt` value needed? Are we not able to use the `submittedAt` value?

aashir21: > Is this `rawSubmittedAt` value needed? Are we not able to use the `submittedAt` value?

The `submitted` field is formatted as "28/05/2025 09:09:42", which is not a valid JS date. Hence, we pass it down to our new method it will always return `null`.

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -3,17 +3,41 @@
...
+import { getOneMonthAgo, getAWeekBeforeSubmittedTime } from "@/utils/timeOperations";
 
 const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
   const tags = metadata?.tags || [];
   const translations = useTranslations("OverviewTab");
+  const [weekBefore, setWeekBefore] = useState<string | null>(null);
+  
+  const MONTH_AGO = getOneMonthAgo();
+
+  const OTHER_RECENT_RUNS = `/test-runs?testName=${metadata?.testName}&bundle=${metadata?.bundle}&group=${metadata?.group}&package=${metadata?.package}&from=${MONTH_AGO}`;
```
eamansour: `OTHER_RECENT_RUNS` will only return runs with the same group - if the group was automatically set by the framework, this will return no other runs

eamansour: MONTH_AGO is a date, not an ISO string, so this will cause issues with the query

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -3,17 +3,41 @@
...
+      const validatedTime = getAWeekBeforeSubmittedTime(metadata?.rawSubmittedAt!);
+      if (validatedTime !== null) {
+        setWeekBefore(validatedTime);
+      }
+      
+    };
+
+    validateTime();
+
+  },[]);
```
eamansour: `rawSubmittedAt` is missing from the useEffect dependencies

galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -3,17 +3,41 @@
...
 
 const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
   const tags = metadata?.tags || [];
   const translations = useTranslations("OverviewTab");
+  const [weekBefore, setWeekBefore] = useState<string | null>(null);
+  
+  const MONTH_AGO = getOneMonthAgo();
+
+  const OTHER_RECENT_RUNS = `/test-runs?testName=${metadata?.testName}&bundle=${metadata?.bundle}&group=${metadata?.group}&package=${metadata?.package}&from=${MONTH_AGO}`;
+  const RETRIES_FOR_THIS_TEST_RUN = `/test-runs?submissionId=${metadata?.submissionId}&from=${weekBefore}`;
```
eamansour: If we click on either of these links, we need to then manually clear the search criteria if we want to search for other runs, which can be cumbersome.

Wonder if we need a "Reset all"/"Clear all" button to clear out all search criteria in a single click

galasa-ui/src/utils/timeOperations.ts
```diff
@@ -173,6 +173,44 @@ export function getYesterday(): Date {
...
+  if(isNaN(submittedDate.getTime())) {
+    result = null;
+  } else {
+
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 7); 
+
+    // Reset time to midnight
+    yesterday.setHours(0, 0, 0, 0); 
+    result = yesterday.toISOString();
```
eamansour: Minor - this is no longer `yesterday`, maybe `lastWeek`? :)

eamansour (approved):

PR #133, opened 2025-07-01
Implement table design tab with drag-and-drop column customization

### LINKED ISSUE #2243: Web UI: Test Runs table columns can be re-ordered and hidden
## Story
As a Galasa user, I want to be able to customize the table of test runs that appear in the web UI's "Test Runs" page, so that I can hide and re-order columns based on my own preferences.

## Background
This story assumes the "Test Runs" page has been created and a populated table of test runs appears on the page. See below for the UI mockup for this story:

Image

## Tasks
- [ ] Add a "Table Design" tab to the test runs tab menu
- [ ] Add a table as shown in the design with rows that correspond to columns on the "Test Runs" page
- [ ] Make the "Table Design" table rows re-orderable using handles or clickable arrows that move a row up/down by one row
- [ ] Add a column to the "Table Design" table to control whether a column on the "Test Runs" page will be shown or hidden (controlled by a checkbox)
- [ ] Set a default selection of columns that are set to be shown (i.e. their "Show column in results" checkbox is checked)
- [ ] Make the table design alter the test runs page URL so that the table design can be shared with the custom query
- [ ] Unit tests for this functionality

Tasneemmohammed0:
## Why
Closes  [galasa-dev/projectmanagement#2243](https://github.com/galasa-dev/projectmanagement/issues/2243)
## Changes
This pull request implements Table Design tab feature that allows users to customize the test runs table by controlling column visibility, ordering, and persistence through URL state management.

**Architecture**
- `TableDesignContent.tsx` - Main table design interface
- `TableDesignRow.tsx` - Individual draggable row component
- Enhanced `TestRunsTabs.tsx` - URL state management integration
- Updated `TestRunsTable.tsx` - Dynamic column rendering

**Features Added**
- Implemented drag-and-drop functionality for column reordering using `@dnd-kit` library
- Added checkbox controls for showing/hiding columns
- Automatically saves state changes to URL and retrieve initial state from the URL.
- Up/down arrow buttons for precise column positioning
- Real-time sync between Table Design tab and Results table


galasa-ui/src/components/test-runs/TableDesignContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+      setSelectedRowIds(rowIds);
+    } else {
+      setSelectedRowIds(allSelected ? [] : rowIds); 
+    }
+  };
+
+  const getRowPosition = (id: string) => tableRows.findIndex(row => row.id === id);
+
+
+  const handleDragEnd = (event: any) => {
```
eamansour: Minor - can this event be typed?

galasa-ui/src/components/test-runs/TableDesignContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+          <div className={styles.cellDragHandle}>
+            <strong>{translations("dragAndDropHeader")}</strong>
+          </div>
+          <div className={styles.cellCheckbox}>
+            <Checkbox
+              aria-label="Select all rows" 
+              id="checkbox-all"
+              checked={selectedRowIds.length === tableRows.length}
+              onChange={handleSelectAll}
+            />
```
eamansour: I think it would be good if the column of checkboxes has some sort of label that mentions what the checkboxes do. At the moment, it's not clear that they can be used to show/hide columns from the results table

galasa-ui/messages/en.json
```diff
@@ -359,5 +359,26 @@
     "remove": "Remove",
     "cancel": "Cancel",
     "save": "Save"
+  },
+  "TableDesignContent": {
+    "description": "Customize your table view by showing, hiding, and reordering columns.",
```
eamansour: I think "your table view" should be clarified a bit to make it clear that the table design tab customizes the table in the results tab - maybe something like "Customize the test run results table by showing, hiding, and reordering columns."? Feel free to come up with something better!

galasa-ui/src/components/test-runs/TableDesignRow.tsx
```diff
@@ -0,0 +1,91 @@
...
+    transform,
+    transition,
+    isDragging,
+  } = useSortable({ id: rowId }); 
+
+  const style = {
+    transform: CSS.Transform.toString(transform),
+    transition,
+    opacity: isDragging ? 0.5 : 1, 
+  };
```
eamansour: Instead of inline styling, we could use CSS classes for the different states and apply them conditionally?

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -23,14 +27,56 @@ interface TestRunsTabProps {
...
-    This page is under construction. In future, you will be able to choose which columns are visible and their order.
-</p>;
-
 
 export default function TestRunsTabs({runsListPromise, requestorNamesPromise, resultsNamesPromise}: TestRunsTabProps) {
   const translations = useTranslations("TestRunsTabs");
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+  const [selectedVisibleColumns, setSelectedVisibleColumns] = useState<string[]>(["submittedAt", "testRunName", "requestor", "testName", "status", "result"]);
```
eamansour: Minor - it might be worth having some kind of const or enum that stores the column names so that we don't have to worry about misspelling them whenever we need to use them

galasa-ui/src/components/test-runs/TableDesignContent.tsx
eamansour: Also, would it be worth displaying a warning message when all columns are hidden? Or should we prevent users from being able to hide all?

eamansour (commented):
The code looks good overall! Just a few comments around the expected behaviour and clarifying labels

galasa-ui/src/components/test-runs/TableDesignRow.tsx
Tasneemmohammed0: Fixed - I've updated the component to use CSS classes for everything.
The only thing I kept inline is the transform style from `dnd-kit`, since it's required for the drag-and-drop animations.

eamansour (commented):
Looks good! I've thought of one other issue that is very similar to https://github.com/galasa-dev/projectmanagement/issues/2275.

When navigating back from the results table using the "Test Runs" link in the breadcrumb, the custom table design gets lost and you would have to re-design the table again. Happy for this to be addressed as part of the above linked issue since I think it's because the query isn't saved when you navigate back at the moment.

eamansour (approved):

PR #134, opened 2025-07-02
fix(results table): correct pagination text in test runs table

### LINKED ISSUE #2306: Web UI: Query results in 99 test runs. Still says page 1 of 1, rather than page 1 of 10...
## Describe the bug
- Using the web UI: Query a list of 99 items or more, going back a few weeks perhaps.
- Look at the results table
- Table says 'page x of x' but the value of x is always the same as x.
  - eg: page 1 of 1. Page 2 of 2. 
- Should be: page 1 of 10, page 2 of 10...

Tasneemmohammed0:
## Why?
Fixes [galasa-dev/projectmanagement#2306](https://github.com/galasa-dev/projectmanagement/issues/2306)


eamansour (approved):
Looks good! Thanks for fixing this!

PR #135, opened 2025-07-02
Improved Search Criteria Behavior

### LINKED ISSUE #2301: Web UI: Search criteria improvements
## Story
As a Galasa user, I want to only save/reset my criteria changes if I have made changes, so that I can easily tell that there are unsaved changes and avoid confusion about clicking "save" or "reset" multiple times.

As a Galasa user, I want the "Cancel" button in the search criteria to be "Reset", so that it's clear that clicking that button will reset all criteria back to their last-saved values.

## Background
This story covers some additional improvements that can be made to the "Search Criteria" tab.

1. The "Save" button should be greyed-out while no changes have been made, and only enabled if there are changes that need saving. 
2. The "Cancel" button should be renamed to "Reset"
3. The "Reset" button should be greyed out while there are no changes to reset
Image

4. The "All" checkbox should be slightly separated away from the other checkboxes so that it is clearly shown to be a separate control, not another value:
Image

5. The test status values should be re-ordered to match the statuses in a test run's lifecycle:
Image

## Tasks
- [x] The "Save" button should be greyed-out while no changes have been made, and only enabled if there are changes that need saving
- [x] The "Reset" button should be greyed out while there are no changes to reset
- [x] Rename the "Cancel" button to "Reset"
- [x] Out-dent the "All" checkbox slightly
- [x] Re-order the checkboxes in the "status" criteria to match the order of statuses in the test run lifecycle

Tasneemmohammed0:
## Why
Closes  [galasa-dev/projectmanagement#2301](https://github.com/galasa-dev/projectmanagement/issues/2301)

## Changes
- **Save** and **Reset** buttons are now disabled when no changes have been made, preventing unnecessary actions.
- Adjusted the indentation of the **"All"** checkbox for better visual alignment.
- Reordered the **status checkboxes** to reflect the logical flow of the test run lifecycle.
- Improved keyboard accessibility by adding full **arrow key navigation** support to the **requestor suggestions** dropdown.

### DIFF
--- a/galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
+++ b/galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
@@ -122,12 +122,14 @@ export default function SearchCriteriaContent({requestorNamesPromise, resultsNam
 
     // Update the local UI state to match the newly selected filter's saved value
     setCurrentInputValue(savedValue);
+
+    const splitSavedValue = savedValue ? savedValue.split(',') : [];
     if (field.id === 'result') {
-      setSelectedResults(savedValue ? savedValue.split(',') : []);
+      setSelectedResults(splitSavedValue);
     } else if (field.id === 'status') {
-      setSelectedStatuses(savedValue ? savedValue.split(',') : []);
+      setSelectedStatuses(splitSavedValue);
     } else if (field.id === 'tags') {
-      setSelectedTags(savedValue ? savedValue.split(',') : []);
+      setSelectedTags(splitSavedValue);
     }
   };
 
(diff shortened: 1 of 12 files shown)


galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
```diff
@@ -110,24 +109,27 @@ export default function SearchCriteriaContent({requestorNamesPromise, resultsNam
...
+    const savedValue = query.get(field.id) || '';
+
+    // Update the local UI state to match the newly selected filter's saved value
+    setCurrentInputValue(savedValue);
+    if (field.id === 'result') {
+      setSelectedResults(savedValue ? savedValue.split(',') : []);
+    } else if (field.id === 'status') {
+      setSelectedStatuses(savedValue ? savedValue.split(',') : []);
+    } else if (field.id === 'tags') {
+      setSelectedTags(savedValue ? savedValue.split(',') : []);
```
eamansour: Very minor - to reduce the duplicate condition here, we can set a const with the split value/empty list before this if-else-if blocks, then use that const directly in the setter methods?

galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
```diff
@@ -205,6 +190,40 @@ export default function SearchCriteriaContent({requestorNamesPromise, resultsNam
...
+      const currentStatuses = [...selectedStatuses].sort();
+      finalValue = JSON.stringify(savedStatuses) === JSON.stringify(currentStatuses);
+      break;
+    }
+
+    case 'tags': {
+      const savedTags = savedValue ? savedValue.split(',').sort() : [];
+      const currentTags = [...selectedTags].sort();
+      finalValue = JSON.stringify(savedTags) === JSON.stringify(currentTags);
+      break;
```
eamansour: Same comment here regarding the duplication of `savedValue ? savedValue.split(',').sort() : []`

galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
```diff
@@ -205,6 +190,40 @@ export default function SearchCriteriaContent({requestorNamesPromise, resultsNam
     }
   };
 
+  const isSaveAndResetDisabled: boolean = (() => {
+    // Get saved value from the query and compare it with the current input value
+    const savedValue = query.get(selectedFilter.id) || '';
+    let finalValue = false;
```
eamansour: Minor - `finalValue` could be clarified to something like `isDisabled` or `isEnabled`?

eamansour (approved):
A few minor comments but otherwise happy to approve!

eamansour (approved):

PR #136, opened 2025-07-03
OverviewTab now displays package name

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2300
- REST API will have to support returning a stream name while fetching test run details, so omitted in this story. Will craft a new story for that.

### DIFF
--- a/galasa-ui/src/components/test-runs/OverviewTab.tsx
+++ b/galasa-ui/src/components/test-runs/OverviewTab.tsx
@@ -49,6 +49,10 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
         title={`${translations("test")}:`}
         value={metadata?.testName}
       />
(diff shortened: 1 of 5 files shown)


galasa-ui/src/components/test-runs/OverviewTab.tsx
```diff
@@ -49,6 +49,10 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
         title={`${translations("test")}:`}
         value={metadata?.testName}
       />
+      <InlineText
+        title={`${translations("package")}:`}
+        value={metadata?.package}
+      />
```
eamansour, line 55: Minor - might be worth updating a unit test to check that the package is being rendered in the overview tab?

eamansour (approved):

PR #137, opened 2025-07-03
Persisting search queries during Test Runs navigation

### LINKED ISSUE #2275: Web UI: Navigating back from the "Test Runs" page should remember the last query
## Story
As a Galasa user, when I click the "Test Runs" breadcrumb link from a test run's details page, I want to return to the "Results" tab showing the results of my last query, so that I don't need to re-enter my custom test runs query each time.

## Background
Currently, when users click through to a specific test run's details page and return to the parent "Test Runs" page, the query that they used previously is lost. This forces users into re-entering their query, which can be cumbersome if their query is very specific.

Users should be able to return to the results page that they saw before clicking into a test run.

There should also be a way for users to completely reset their query if they wish to do so at any point.

Tasneemmohammed0:
## Why?
Closes [galasa-dev/projectmanagement#2275](https://github.com/galasa-dev/projectmanagement/issues/2275)
## Changes 
- The search query is now stored in `sessionStorage` before redirecting to the Test Run Page.
- The Single Test Run page retrieves the last saved query from `sessionStorage` (stored in state).
- Added unit test to verify the functionality.
   
> [!NOTE]
> Why `sessionStorage` instead of URL parameters? 
> Passing the query via URL is intentionally avoided to maintain a clean URL structure for the Single Test Run page. This ensures the URL remains available for test-run-specific qualifiers without cluttering it with unnecessary search parameters.

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunDetails.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunDetails.tsx
@@ -22,6 +22,7 @@ import { HOME, TEST_RUNS } from '@/utils/constants/breadcrumb';
 import TestRunSkeleton from './TestRunSkeleton';
 import { useTranslations } from 'next-intl';
 import StatusIndicator from '../common/StatusIndicator';
+import { useRouter, useSearchParams } from 'next/navigation';
 
 interface TestRunDetailsProps {
   runId: string;
@@ -32,6 +33,9 @@ interface TestRunDetailsProps {
 
 // Type the props directly on the function's parameter
 const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsPromise }: TestRunDetailsProps) => {
+  const searchParams = useSearchParams();
+  const router = useRouter();
+  const translations = useTranslations("TestRunDetails");
 
   const [run, setRun] = useState<RunMetadata>();
   const [methods, setMethods] = useState<TestMethod[]>([]);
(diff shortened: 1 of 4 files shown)


galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -29,12 +29,13 @@ import { TableHeadProps } from "@carbon/react/lib/components/DataTable/TableHead
...
 import { useEffect, useMemo, useState } from "react";
-import { useRouter } from "next/navigation";
+import { useRouter, useSearchParams } from "next/navigation";
 import ErrorPage from "@/app/error/page";
 import { TestRunsData } from "@/utils/testRuns";
-import {MAX_RECORDS} from "@/utils/constants/common";
+import { MAX_RECORDS} from "@/utils/constants/common";
 import { useTranslations } from "next-intl";
 import { InlineNotification } from "@carbon/react";
+import { run } from "node:test";
```
eamansour: I think this node:test import can be removed

galasa-ui/src/tests/components/TestRunDetails.test.tsx
```diff
@@ -6,6 +6,7 @@
 import React from 'react';
 import { render, screen, act } from '@testing-library/react';
 import TestRunDetails from '@/components/test-runs/TestRunDetails';
+import { get } from 'http';
```
eamansour: This import doesn't seem to be used, would be good to remove if so

galasa-ui/src/components/test-runs/TestRunDetails.tsx
```diff
@@ -96,11 +107,14 @@ const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsP
   if (isError) {
     return <ErrorPage />;
   }
-  
+
+  // Build the test runs route with the saved query string
+  const testRunsRoute = savedQuery ? `${TEST_RUNS.route}?${savedQuery}` : TEST_RUNS.route;
+  const testRunsBreadCrumb = { ...TEST_RUNS, route: testRunsRoute };
```
eamansour, line 113: I think this breadcrumb will still take you back to the "timeframe" tab, rather than the "results" tab. I'll raise a separate issue for that!

eamansour (approved):
Looks good, thanks for sorting this out!

PR #138, opened 2025-07-03
hotfix(search-criteria): resolve duplicate checkbox IDs and style issues

### LINKED ISSUE #2301: Web UI: Search criteria improvements

Tasneemmohammed0:
## Why?
Fixes https://github.com/galasa-dev/projectmanagement/issues/2301#issuecomment-3031626653

### DIFF
--- a/galasa-ui/src/components/test-runs/CustomCheckBoxList.tsx
+++ b/galasa-ui/src/components/test-runs/CustomCheckBoxList.tsx
@@ -59,7 +59,7 @@ export default function CheckBoxList({ title, items, selectedItems, onChange, on
         className={styles.checkBoxList}
       >
         <Checkbox
-          id="checkbox-all"
+          id="status-checkbox-all"
           labelText="All"
           onChange={(event: React.ChangeEvent<HTMLInputElement>) => handleAllChange(event.target.checked)} 
           checked={areAllSelected}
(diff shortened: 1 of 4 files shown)


eamansour (approved):

galasa-ui/src/components/test-runs/CustomCheckBoxList.tsx
```diff
@@ -52,14 +52,15 @@ export default function CheckBoxList({ title, items, selectedItems, onChange, on
     onChange(checked ? items : []);
   };
   
+  console.log("Disable Save and Reset:", disableSaveAndReset);
```
eamansour: I think we can remove this log statement

PR #139, opened 2025-07-04
  fix: correct tab routing when navigating from test run details

### LINKED ISSUE #2313: Web UI: Test runs breadcrumb and Test run links should take you back to the "Results" tab rather than the "Timeframe" tab
## Describe the bug
When clicking on the "Test runs" breadcrumb link from a test run's details panel, the link takes you back to the "Timeframe" tab rather than the "Results" tab.

Similarly, when clicking on either the "View other recent runs of the same test" or the "View retries of this test run", the links also take you to the "Timeframe" tab.

## Steps to reproduce
1. Go to a test run's details page
2. Click on the "Test runs" breadcrumb link
3. Notice that you will be taken to the "Timeframe" tab

## Expected behavior
The links should take you to the "Results" tab.

Tasneemmohammed0:
## Why? 
Fixes [https://github.com/galasa-dev/projectmanagement/issues/2313](https://github.com/galasa-dev/projectmanagement/issues/2313)

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunsTabs.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunsTabs.tsx
@@ -13,12 +13,12 @@ import TableDesignContent from './TableDesignContent';
 import { usePathname, useSearchParams, useRouter } from "next/navigation";
 import { TestRunsData } from "@/utils/testRuns";
 import { useTranslations } from "next-intl";
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { RESULTS_TABLE_COLUMNS, COLUMNS_IDS} from '@/utils/constants/common';
 
 interface TabConfig {
+  id: string;
   label: string;
-  component: React.ReactNode;
 }
 
 interface TestRunsTabProps {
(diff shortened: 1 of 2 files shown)


galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -27,106 +27,123 @@ interface TestRunsTabProps {
...
-  // Load from URL parameters first (only on mount)
-  useEffect(() => {
-    const visibleColumnsParam = searchParams.get("visibleColumns");
-    const columnsOrderParam = searchParams.get("columnsOrder");
+  // Initialize selectedIndex based on URL parameters or default to first tab
+  const [selectedIndex, setSelectedIndex] = useState(() => {
+    const tabParam = searchParams.get("tab");
+    const initialIndex = tabParam ? TABS_IDS.indexOf(tabParam) : -1;
+    return initialIndex !== -1 ? initialIndex : 0;
+  });
```
eamansour, line 43: Instead of indexes, would it be easier to track the selected tab by its ID?

So instead of a list of tabs, we could use a map of key-value pairs where the keys are the IDs (e.g. 'timeframe', 'table-design' etc, and the values are the tab labels. Just a thought!

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -27,106 +27,123 @@ interface TestRunsTabProps {
...
+  const [isInitialized, setIsInitialized] = useState(false);
+  useEffect(() => {setIsInitialized(true);}, []);
+
+  // Define the tabs with their corresponding labels
+  const TABS_CONFIG: TabConfig[] = [
+    { id: TABS_IDS[0], label: translations('tabs.timeframe') },
+    { id: TABS_IDS[1], label: translations('tabs.tableDesign') },
+    { id: TABS_IDS[2], label: translations('tabs.searchCriteria') },
+    { id: TABS_IDS[3], label: translations('tabs.results') },
+  ];
```
eamansour, line 82: If we used a map for the tabs, then I don't think we would need this list either

eamansour (approved):
Looks good, just one suggestion but happy to approve

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
Tasneemmohammed0: Good idea! However, we need to track the actual index so we can use it with the Carbon `Tabs` component for proper control.

eamansour: No problem, happy to keep it as-is in that case!

PR #140, opened 2025-07-05
Optimize Test Run Page: No Unnecessary Fetches on URL Changes

Tasneemmohammed0:
## Why? 
Refer to  [https://github.com/galasa-dev/projectmanagement/issues/2310](https://github.com/galasa-dev/projectmanagement/issues/2310)

We’ve reworked how data is fetched on the Test Run page. Instead of relying on server-side fetching tied to URL changes, we now use client-side caching and state management with `@tanstack/react-query`.

### What Was the Problem?
Previously, we fetched data on the server every time the URL changed. This tightly coupled the app’s data fetching to the URL, causing some issues:
- We had to change the URL whenever search criteria were updated, just to trigger a data fetch. Adding simple guards like `if` statements to skip unnecessary fetches wasn’t effective.
- Many of the URL parameters ( like `orderColumns` or `visibleColumns`) are purely related to the UI and don't trigger server requests, the old approach would refetch on every URL change, even for minor UI adjustments.

### What's New?
- React Query introduces a dedicated client-side cache. The main key of this implementation is a *canonical query key*, which is constructed from the URL.
  - It separates data-driving parameters from UI-only parameters.
  - It normalizs order-independent parameters (`tags=A, B` is treated the same as `tags=B, A`).
  - A network request is **only triggered** when the underlying data query genuinely changes. Toggling column visibility or reordering columns or changing tabs will **never** cause a refetch.
- The responsibility of managing loading, error, and data states is abstracted away by the `useQuery` hook. 
- Configured `useQuery` with `enabled: true` only when the "Results" tab is active, preventing fetches for hidden content.

I’ve also added unit tests to cover this new behavior, ensuring that data fetching works as expected and that no requests are triggered on irrelevant URL changes or tab switches.

> [!NOTE]
> A common workaround for similar cases is to use `window.history.replaceState` to change the URL without triggering a fetch. However, we’ve intentionally avoided this approach. It breaks Next.js’s routing and can cause the app’s state to get out of sync with the router.


galasa-ui/src/utils/constants/common.ts
```diff
@@ -55,9 +55,26 @@ const RESULTS_TABLE_COLUMNS = [
   { id: "result", columnName: "Result" },
 ];
   
+const PARAMS = {
```
eamansour: Very minor - instead of `PARAMS`, this could be more descriptive and be something like `RUN_QUERY_PARAMS`?

Feel free to come up with something better!

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -90,18 +90,66 @@ export default function TestRunsTabs({runsListPromise, requestorNamesPromise, re
...
-  }, [selectedVisibleColumns, columnsOrder, isInitialized, pathname, router, searchParams, selectedIndex]);
+  }, [selectedVisibleColumns, columnsOrder, isInitialized, pathname, router, selectedIndex]);
 
   const handleTabChange = (event: {selectedIndex : number}) => {
     const currentIndex = event.selectedIndex;
     setSelectedIndex(currentIndex);
   };
 
+  // Create a canonical query key based on the search parameters so that it won't refetch unnecessarily
+  const queryKey = useMemo(() => {
```
eamansour, line 106: Just tried these changes locally and while they do seem to make an improvement, when selecting "All" from the "results" and "status" criteria, no test runs are returned even though there are runs that should be returned.

eamansour: Also noticed an issue when adding tags to the criteria:

If you add a tag that doesn't exist, then save the tags and go to results - no runs are found (this is fine). However, if you go back and add a tag that does exist and should return runs, then no runs are returned.

Similarly, if you add a tag that does exist, then save and go to results - runs will be returned (which is good!). However, if you remove that tag and add a tag that doesn't exist, the runs from the previous results are still shown, which isn't right!

eamansour: Haven't tried it with other search criteria, but this is worth investigating to make sure runs are returned as expected.

Might be worth adding a unit test that performs a couple of interactions, like adding a tag and then removing it, to check that we're sending a request to the API server when we expect it to.

eamansour (commented):
Looking good - just a few issues I've noticed when trying the changes out locally which would be great to iron out!

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
Tasneemmohammed0: > Just tried these changes locally and while they do seem to make an improvement, when selecting "All" from the "results" and "status" criteria, no test runs are returned even though there are runs that should be returned.

Thanks for spotting this, @eamansour! It's a small fix related to #135 . Turns out the "Allocated," "Waiting," and "Submitted" statuses weren't available as query parameters yet, which caused the API to throw an error.

Tasneemmohammed0: > Also noticed an issue when adding tags to the criteria:
> 
> If you add a tag that doesn't exist, then save the tags and go to results - no runs are found (this is fine). However, if you go back and add a tag that does exist and should return runs, then no runs are returned.
> 
> Similarly, if you add a tag that does exist, then save and go to results - runs will be returned (which is good!). However, if you remove that tag and add a tag that doesn't exist, the runs from the previous results are still shown, which isn't right!

Fixed. Was a small typo of not including "Tags" in the relevant params.

eamansour (approved):
Thanks for sorting out those small issues, happy to approve!

PR #141, opened 2025-07-07
Implement "Clear Filters" functionality

### LINKED ISSUE #2308: Web UI: Add a control to reset the search criteria that are currently set
## Story
As a Galasa user, I want to quickly be able to reset all search criteria to their default values, so that I don't have to manually clear out the search criteria I've set.

## Background
Currently, in order to reset all search criteria for the test runs table, you would have to manually clear out each criteria input field and then save the changes. Alternatively, you could click on the "Test runs" option in the menu bar at the top and that will clear out the criteria as well.

We should add some kind of "Reset to defaults" button to one or more of the "Test run" tab panels for users to be able to quickly clear out all currently-set search criteria and set them back to their default values.

This control can be added to the "Search Criteria" tab so that users can quickly re-do their search criteria settings.

## Tasks
- [x] Add a "Reset to defaults" button to the "Search Criteria" tab that clears out the search criteria for the results table

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2308](https://github.com/galasa-dev/projectmanagement/issues/2308)


galasa-ui/src/styles/TestRunsPage.module.css
```diff
@@ -128,6 +128,12 @@
     justify-content: flex-end;
 }
 
+.resetToDefaultsButton {
+  position: absolute;
+  bottom: 3rem;
+  right: 5rem;
+}
```
eamansour: For smaller screens, users will have to scroll down to see this button.

Would it be better to have this button at the top right? Maybe having the button in-line with the tabs or just below would be better?

eamansour (approved):
One comment but otherwise looks good, happy to approve!

galasa-ui/src/components/test-runs/SearchCriteriaContent.tsx
```diff
@@ -315,6 +327,14 @@ export default function SearchCriteriaContent({requestorNamesPromise, resultsNam
...
         {renderComponent(selectedFilter)}
       </div>
+      <Button 
+        type="button"
+        kind="secondary"
+        className={styles.resetToDefaultsButton}
+        onClick={handleResetToDefaults}
+      >
+        {translations("clearFilters")}
+      </Button>
```
eamansour: Would it be worth setting this button to disabled if there are no criteria that have been set? It can then be enabled once there are criteria that can be cleared.

That would avoid confusion if someone wants to clear their criteria and notices that the "Clear Filters" button is still clickable.

PR #144, opened 2025-07-08
refactor(theme):removed experimental flag,enable theme by default

### LINKED ISSUE #2283: Web UI: Support Light/Dark mode
# Story As a galasa web UI user, I want to see the UI content in different themes.

# Background

In settings panel we need this: 

```
Color Theme
===========
o Light mode
o Dark mode 
o Light or dark to match my browser preferences
o Light or dark to match my desktop preference
```

With a radio button. 

There is an IBM Carbon theme tag/component which can pass down the theme to the child components. We can use that?

Can this (please) be stored in browser local storage ?

# Tasks
- [x] Needs hiding behind a feature-flag initially "Control color theme". 
- [x] If the feature flag is turned off, revert to the light mode we use today in all our panels.
- [x] Settings page needs a toggle button with the available options
- [x] All panels need re-working to support a theme control to control the fonts and colour palettes.

shivambisen:
##  Summary

This PR removes the experimental feature flag for the theme system and enables it as a default feature across the application.

---
Closes https://github.com/galasa-dev/projectmanagement/issues/2283

## Changes Made

- Removed `FEATURE_FLAGS.THEME` from the codebase.
- Enabled the theme functionality unconditionally.
- Cleaned up conditional checks that previously depended on the theme flag.
- Updated tests to reflect the removal of the flag (where applicable).


galasa-ui/src/tests/components/PageHeaderMenu.test.tsx
```diff
@@ -158,6 +158,23 @@ test('clicking log out button calls handleDeleteCookieApiOperation, RESPONSE OK'
...
+    writable: true,
+    value: jest.fn().mockImplementation(query => ({
+      matches: false, // or true, depending on your test case
+      media: query,
+      onchange: null,
+      addListener: jest.fn(), // deprecated
+      removeListener: jest.fn(), // deprecated
+      addEventListener: jest.fn(), // modern
+      removeEventListener: jest.fn(),
+      dispatchEvent: jest.fn(),
```
eamansour: Do we need to define all of these functions in this mock?

We could only implement the methods/properties that we use perhaps?

galasa-ui/src/tests/components/PageHeader.test.tsx
```diff
@@ -38,6 +38,22 @@ jest.mock('next-intl', () => ({
...
+    writable: true,
+    value: jest.fn().mockImplementation(query => ({
+      matches: false, // or true, depending on your test case
+      media: query,
+      onchange: null,
+      addListener: jest.fn(), // deprecated
+      removeListener: jest.fn(), // deprecated
+      addEventListener: jest.fn(), // modern
+      removeEventListener: jest.fn(),
+      dispatchEvent: jest.fn(),
```
eamansour: Do we need to define all of these functions in this mock?

We could only implement the methods/properties that we use perhaps?

galasa-ui/src/tests/index.test.tsx
```diff
@@ -48,6 +48,23 @@ jest.mock('next/headers', () => ({
...
+    writable: true,
+    value: jest.fn().mockImplementation(query => ({
+      matches: false, // or true, depending on your test case
+      media: query,
+      onchange: null,
+      addListener: jest.fn(), // deprecated
+      removeListener: jest.fn(), // deprecated
+      addEventListener: jest.fn(), // modern
+      removeEventListener: jest.fn(),
+      dispatchEvent: jest.fn(),
```
eamansour: Do we need to define all of these functions in this mock?

We could only implement the methods/properties that we use perhaps?

eamansour (commented):
Looks good overall! Just one minor comment that applies to three different places

galasa-ui/src/tests/components/PageHeaderMenu.test.tsx
shivambisen: Good point  since we're not using any of those matchMedia methods in our tests, it's fine to remove them from the mock.

eamansour (approved):

PR #146, opened 2025-07-09
Implement Breadcrumb History + Navigation Link Improvements

### LINKED ISSUE #2314: Web UI: Improvements to the "View other test runs" links
## Describe the bug
This defect covers a few improvements that can be made to the behaviour of the links used to view other test runs from a test run's details page:

Image

1. The font size of the links is slightly smaller than the text above them
2. When clicking on either link, it doesn't retain the breadcrumbs from the fact that you got there from a specific test run's details page

Raised as a separate issue in https://github.com/galasa-dev/projectmanagement/issues/2313: The links don't take you straight to the Results tab of the query

## Expected behavior
1. The font size should be consistent with the text above the links
2. The links should retain the test run's details page breadcrumb. For example, 'test run: U1234' item in the breadcrumb of the search panel

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2314](https://github.com/galasa-dev/projectmanagement/issues/2314)
This PR addresses breadcrumb usability issues by introducing a dynamic breadcrumb history mechanism, improving navigation throughout the Web UI.
## Changes
- Introduced a custom breadcrumb history (`useBreadCrumbHistory`) hook that:
   - Persists breadcrumb history in **sessionStorage**. 
   - Automatically updates breadcrumbs on navigation and browser back/forward actions.
- Enhanced the `BreadCrumb` component to:
   - Support overflow menus for deep navigation paths.
   - Display breadcrumbs based on dynamic history rather than static definitions.
   
> [!tip]
> Now, adding to the breadcrumb history is simple, just push the desired breadcrumb to the history, and the custom hook handles the rest automatically.


galasa-ui/src/app/test-runs/page.tsx
```diff
@@ -12,13 +12,26 @@ import { Suspense } from "react";
...
+  const fromRunId = searchParams.fromRunId || "";
+  
+  const breadCrumbItems = fromRunName ? [
+    HOME, 
+    {
+      // This is the link back to the originating run
+      title: "testRunName",
+      values: { runName: fromRunName },
+      route: `/test-runs/${fromRunId}`,
+    },
```
eamansour: Should the breadcrumb include the original "Test Runs" link in case someone wants to jump all the way back to where they were before? Like:

```
Home / Test Runs / Test Run: ABC123
```

Instead of:
```
Home / Test Run: ABC123
```

I suppose the breadcrumb could become quite long but I think that's fine because we want to keep track of where the user was previously so that they can jump back however they want, so it could become something like:

```
Home / Test Runs / Test Run: ABC123 / Test Runs / Test Run: DEF456 / Test Runs / ...
```

If the breadcrumb becomes too long, we can add a `...` somewhere in the middle to keep it short? Like:


```
Home / Test Runs / Test Run: ABC123 / ... / Test Run: DEF456
```

eamansour (commented):
Just one comment that I can think of at the moment!

galasa-ui/src/app/test-runs/page.tsx
Tasneemmohammed0: Well, turns out this change shakes up the whole flow! Drafting the PR while I sort out the unit tests. 🙂

galasa-ui/messages/en.json
```diff
@@ -6,7 +6,7 @@
     "home": "Home",
     "users": "Users",
     "testRuns": "Test Runs",
-    "testRunName": "Test Run: {runName}"
+    "testRunName": "Test Run {runName}"
```
eamansour: I think we can remove the "Test Run" bit from the breadcrumb, so the breadcrumb would look like:

```
Home / Test Runs / U1234 / Test Runs / A4567
```

Same change applies to the German translations

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -38,21 +39,65 @@ function BreadCrumb({
     theme = 'g10';
   }
 
+  const renderBreadCrumbItems = (items: BreadCrumbProps[]) => {
+    return items.map((item) => {
+      const translatedTitle = translations(item.title);
+      const displayText = translatedTitle.startsWith("Breadcrumb.") ? item.title : translatedTitle;
+      
+      return (
+        // Use onClick instead of href
```
eamansour: Really minor - this comment mentions onClick but it doesn't look like onClick is being used anywhere. Can this comment be removed/updated?

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -37,19 +39,65 @@ function BreadCrumb({
...
     <Theme theme={theme}>
-      <Breadcrumb className={styles.crumbContainer}>
-        {breadCrumbItems.map((item, idx) => {
-          return (
-            <BreadcrumbItem key={idx} isCurrentPage={false} href={item.route}>
-              {translations(item.title)}
-            </BreadcrumbItem>
-          );
-        })}
+      <Breadcrumb className={styles.crumbContainer} noTrailingSlash>
```
eamansour: I wonder if we should keep the trailing slash in the breadcrumbs since we're not including the current page in the breadcrumb (as a non-clickable element).

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -37,19 +39,65 @@ function BreadCrumb({
...
+        {
+          breadCrumbItems.length <= BREADCRUMB_THRESHOLD ? renderBreadCrumbItems(breadCrumbItems) : 
+            <>
+              {/* Render the first 2 items */}
+              {renderBreadCrumbItems(breadCrumbItems.slice(0, 2))}
```
eamansour, line 84: Would this cause any issues if there's only one breadcrumb link?

Also, would it be worth using a const value that we can use throughout this file in case we ever want more than 2 items being rendered?

galasa-ui/src/hooks/useHistoryBreadCrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+    setItems((prevItems) => {
+      const currentPathIndex = prevItems.findIndex(item => (item.route === fullPath));
+
+      // If the current path is found in our history, truncate the list to that point
+      if (currentPathIndex > -1) {
+        const newItems = prevItems.slice(0, currentPathIndex);
+        sessionStorage.setItem(SESSION_STORAGE_KEY, newItems.length <= 0 ? JSON.stringify([HOME]):  JSON.stringify(newItems));
+        return newItems;
+      }
+      return prevItems;
```
eamansour: Multiple returns here! I think one return at the end would make this code easier to follow

galasa-ui/src/tests/components/BreadCrumb.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+import BreadCrumb from '@/components/common/BreadCrumb';
+
+
+// Mock useTranslations hook to return a mock translation function
+jest.mock('next-intl', () => ({
+  useTranslations: () => (key: string) => {
+    if (key.startsWith("Breadcrumb.")) {
+      return key; // Return the key as is for testing
+    }
+    return key;
```
eamansour: This if statement doesn't seem to be doing anything different - we could simplify this to:
```
return key;
```

galasa-ui/src/hooks/useHistoryBreadCrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }, [pathname, searchParams]);
```
eamansour: Does resetBreadcrumbs need to be in this useEffect's list of dependencies? I think this will show up as a warning in vscode.

galasa-ui/src/tests/components/TestRunsTabs.test.tsx
```diff
@@ -131,6 +130,17 @@ const wrapper = ({ children }: { children: React.ReactNode }) => (
...
 
+const generateMockApiData = (count: number) => {
+  return {
+    limitExceeded: false,
+    runs: Array.from({ length: count }, (_, index) => ({
+      runId: `${index + 1}`, 
+      testStructure: { runName: `Test Run ${index + 1}` },
+    })),
+  };
+};
```
eamansour: Doesn't look like `generateMockApiData` is being used, can it be removed?

galasa-ui/src/tests/components/OverviewTab.test.tsx
```diff
@@ -263,4 +273,14 @@ describe('OverviewTab - Time and Link Logic', () => {
...
     expect(retriesLink).toBeUndefined();
   });
+
+  it('push link bread crumb when any of the links is clicked', () => {
+    render(<OverviewTab metadata={completeMetadata} />);
+
+    const links = screen.getAllByTestId('mock-link');
+    links.forEach(link => {
+      link.click();
+      expect(pushBreadCrumbMock).toHaveBeenCalledWith(expect.anything());
```
eamansour: Is there a specific value we expect this to have been called with?

galasa-ui/src/components/common/BreadCrumb.tsx
```diff
@@ -3,30 +3,32 @@
...
 
 function BreadCrumb({
   breadCrumbItems,
-}: {
-  breadCrumbItems: BreadCrumbProps[];
-}) {
+}: CustomBreadCrumbProps) {
   const translations = useTranslations("Breadcrumb");
   const current = useTheme().theme;
+  const router = useRouter();
```
eamansour, line 27: Doesn't look like we're using `router` anywhere here, can this + the import be removed?

eamansour (commented):
Looks good! Just a few comments - also, spotted some bits of code that aren't being used and I might've missed some more, so it would be good to make sure we're not leaving anything unused!

galasa-ui/src/components/common/BreadCrumb.tsx
Tasneemmohammed0: No, it wouldn’t. If the number of breadcrumb links is below the threshold, it will simply render all breadcrumbs.

galasa-ui/src/components/common/BreadCrumb.tsx
Tasneemmohammed0: It is triggered when an `OverflowMenuItem` is clicked to navigate to the item's route.

eamansour (approved):
Thanks for making those changes, happy to approve :)

PR #149, opened 2025-07-09
Table Design Improvements 

### LINKED ISSUE #2333: Web UI: Table design Improvements
## Describe the bug
1. Sharing a link to the test runs page with a custom table design doesn't include the customised table design, so a different table will be shown. See video example:

https://github.com/user-attachments/assets/da60f2a4-6afb-4a76-a8fe-812470fd4aeb

2. When hiding all columns in the table design, the results panel doesn't display a message indicating that all columns are hidden:

Table design:
Image

Results:

Image

## Expected behavior
1. Sharing a link should keep the same table design regardless of the tab it was sent from
2. The results tab should display a message to inform the user that all columns are hidden

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2333](https://github.com/galasa-dev/projectmanagement/issues/2333)

### DIFF
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -245,6 +245,7 @@
       "default": "Showing test runs submitted in the last 24 hours",
       "range": "Showing test runs submitted between {from} and {to}"
     },
+    "noColumnsSelected": "No columns selected – nothing to display. Please select one or more columns.",
     "pagination": {
       "backwardText": "Previous page",
       "forwardText": "Next page",
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -245,7 +245,7 @@
       "default": "Showing test runs submitted in the last 24 hours",
       "range": "Showing test runs submitted between {from} and {to}"
     },
-    "noColumnsSelected": "No columns selected – nothing to display. Please select one or more columns.",
+    "noColumnsSelected": "All of the columns have been hidden in the table design tab, so no result details will be visible. Please adjust the table design to show the columns that you would like to see.",
     "pagination": {
       "backwardText": "Previous page",
       "forwardText": "Next page",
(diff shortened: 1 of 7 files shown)


galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -183,6 +183,10 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
     router.push(`/test-runs/${runId}`);
   };
 
+  if (visibleColumns.length === 0) {
+    return <p>{translations('noColumnsSelected')}</p>;
+  }
```
eamansour, line 188: I think it would be worth adding a unit test to check that hiding all columns will display this message

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -91,7 +91,12 @@ export default function TestRunsTabs({ requestorNamesPromise, resultsNamesPromis
     
     const params = new URLSearchParams(searchParams.toString());
     params.set(RUN_QUERY_PARAMS.TAB, currentTab.id);
-    params.set(RUN_QUERY_PARAMS.VISIBLE_COLUMNS, visibleColumnsParam);
+    if(selectedVisibleColumns.length > 0) {
+      params.set(RUN_QUERY_PARAMS.VISIBLE_COLUMNS, visibleColumnsParam);
+    } else {
+      // If no columns are selected, we can clear the parameter
+      params.delete(RUN_QUERY_PARAMS.VISIBLE_COLUMNS);
+    }
```
eamansour, line 99: Do we have unit tests that check that this query parameter is only cleared when it should be?

galasa-ui/messages/en.json
```diff
@@ -245,6 +245,7 @@
       "default": "Showing test runs submitted in the last 24 hours",
       "range": "Showing test runs submitted between {from} and {to}"
     },
+    "noColumnsSelected": "No columns selected – nothing to display. Please select one or more columns.",
```
eamansour: This is good but could be clarified to mention the table design tab, maybe something like:

"All of the columns have been hidden in the table design tab, so no result details will be visible. Please adjust the table design to show the columns that you would like to see."

Feel free to come up with a better message!

eamansour (commented):
Looks good - just a few minor comments around unit tests and the message that appears :)

eamansour (approved):
Thanks for making those changes!

PR #150, opened 2025-07-10
Add Share Button Functionality with Clipboard Copy

### LINKED ISSUE #2330: Web UI: Simple share of the URL using clipboard
## Story
As a Galasa tester using the web UI, I want to see a share icon on the test run page header, and the test runs query page header, so that I can easily share the URL with others to get to the same place as me.

## Background
Something like this icon: 

Image

In this location:
Image

Initially we can just say in a message which disappears after a bit "The link to this page has been copied to your clipboard".

Later on we can think about email ...etc.

## Tasks
- [x] Question: Is that the best icon for this ? Or should we use more of a copy-to-clipboard icon if there is one ?
- [x] Icon visible and working on the single Test Run panel (at all times, regardless of tab being used, as it's inside the header)
- [x] Icon visible and working on the Test Runs search panel (at all times, regardless of tab being used, as it's inside the header)
- [x] Hover-help when someone lingers over the button
- [x] Unit testing

shivambisen:
##  Closes:
Closes [#2335](https://github.com/galasa-dev/projectmanagement/issues/2330)

---

##  Description:
- Added a share icon button to the Test Run panels.
- Implemented clipboard copy functionality using `navigator.clipboard.writeText`.
- Tooltip included for better user experience on hover.
- Introduced a unit test to verify that clicking the share button copies the current page URL.

demo-

https://github.com/user-attachments/assets/99fcbae4-0959-4d6f-b502-997a095f5f92

### DIFF
--- a/galasa-ui/messages/de.json
+++ b/galasa-ui/messages/de.json
@@ -205,6 +205,8 @@
   "TestRunDetails": {
     "title": "Testlauf: {runName}",
     "status": "Status",
+    "copymessage": "Seiten-URL kopieren",
+    "copiedmessage": "Kopiert",
     "result": "Ergebnis",
     "test": "Test",
     "tabs": {
--- a/galasa-ui/messages/en.json
+++ b/galasa-ui/messages/en.json
@@ -204,6 +204,8 @@
   },
   "TestRunDetails": {
     "title": "Test Run: {runName}",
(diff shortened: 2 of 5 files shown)


techcobweb (approved):

techcobweb:
Code looks good to me, but conflicts need resolving before we can merge pls.

PR #152, opened 2025-07-14
Add Multi-Level Sort Support with Shareable URL State

### LINKED ISSUE #2286: Web UI: User can specify the sort order of fields
## Story
As a Galasa web UI user when I am querying historic test runs, I want to be able to specify the sort order of fields so that the search results appear in the order I want

## Background
Not got a design for where this should be placed or how. Not on the table though.

It should be either in the table design, or in the search criteria panels I think.

## Tasks
- [ ] Design and agree how this gets specified
- [ ] Get the UI design reviewed by the rest of the team, and possibly some UI designer if we can get their time.
- [ ] Implement
- [ ] unit test

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2286](https://github.com/galasa-dev/projectmanagement/issues/2286)
## Changes 
- Introduced a new *"Sort Order"* column in the results table to enable multi-level sorting based on user-defined priority.
- Synced the sort state with the URL, making it shareable and bookmarkable.
- Added unit tests covering multiple sorting scenarios.

### DIFF
--- a/galasa-ui/messages/de.json
+++ b/galasa-ui/messages/de.json
@@ -376,7 +376,8 @@
     "columnName": "Spaltenname",
     "showHideHeader": "Spalten ein-/ausblenden",
     "warning": "Warnung",
-    "noColumnsSelected": "Keine Spalten ausgewählt – nichts anzuzeigen. Bitte wählen Sie eine oder mehrere Spalten aus"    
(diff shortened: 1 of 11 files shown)


techcobweb (approved):
These changes look good to me.

PR #153, opened 2025-07-14
hotfix: correct date field comparison in multi-level sort

Tasneemmohammed0:


galasa-ui/src/tests/components/TestRunsTabs.test.tsx
```diff
@@ -659,6 +659,44 @@ describe('TestRunsTabs Component', () => {
       });
     });
 
+    test('sorts by date in ascending order', async () => {
```
techcobweb, line 662: Shouldn't the path to the test file roughly match the path to the implementation file ?
eg: galasa-ui/src/tests/components/test-runs/TestRunsTabs.test.tsx ?

If a folder for the source code helps organise the product code into natural logical pieces, the same segregation should be used for test code also ?

techcobweb (approved):

galasa-ui/src/tests/components/TestRunsTabs.test.tsx
Tasneemmohammed0: You're right, I totally agree with that approach. When I first started working in the codebase, this was the existing structure, so I kept it as-is. Refactoring the test files to mirror the implementation structure would definitely be a good improvement, but it might be better suited for a separate PR since it would involve changing paths for many components which is a major code refactor.

techcobweb: Ok. Lets merge this one in and do that in a separate PR as you say.

PR #159, opened 2025-07-15
Shorten URL by mapping keys/values and compressing state

### LINKED ISSUE #2278: Web UI: "Test Runs" query URL should be shortened
## Story
As a Galasa user searching for test runs, I want the results' URL to be short so that I can easily share the URL without accidentally missing any details and avoid reaching any URL length limits for my browser.

## Background
When querying for test runs, the timeframe and search criteria are added to the web UI URL, which can become very long the more specific a query is.

We should shorten the query URL to make it easier to share and avoid running into any browser limits.

## Tasks
- [ ] ...

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2278](https://github.com/galasa-dev/projectmanagement/issues/2278)
## Changes
**Architecture**
- `urlEncoder`
  - Compresses/decompresses the full URL state object using the [`lz-string`](https://pieroxy.net/blog/pages/lz-string/index.html) library.
- `urlStateMappers`
  - Maps verbose keys/values to compact short forms (e.g. submittedAt → sA, Passed → P) to make the URL even more shorter, as even when using [`lz-string`](https://pieroxy.net/blog/pages/lz-string/index.html) to compress, our URL has alot of filters to carry - so it's still long.
  - Reversible mapping logic using `minifyState()` and `expandState()`.
- Refactor `TestRunsTabs` Component
  - Centralized state management for child tabs.
  - Replaces long query strings with a compressed `q` parameter.
  - Supports decoding back into full UI state on page load.
  
  Here is a graph that shows more how the flow goes on:
  
image


galasa-ui/package.json
```diff
@@ -25,6 +25,7 @@
     "markdown-it": "^14.1.0",
     "next": "^14.0.3",
     "next-intl": "^4.1.0",
+    "pako": "^2.1.0",
```
techcobweb: Pako has a zlib license for part of it. Are you pulling that part in ?
The rest is covered by an MIT license which is probably OK for us to rely on.

Tasneemmohammed0: I've removed this dependency and switched to using `lz-string`.

techcobweb:
Because this URL shortening is non-trivial, I think we need somewhere which describes the process, with examples at every stage.

Like an in and .md file concerning the shortening of URLs, or in the urlEncoder code file itself inside a comment.

eg: 
This query from the tabs:

Gets mapped to this string:

Which gets lz-string compressed into this string: 

And the reverse happens when the short URL is decoded.

techcobweb:
If we wanted to translate "Passed" and "Failed" ...etc. into German before we displayed them, how would this effect this code ?

Tasneemmohammed0:
> If we wanted to translate "Passed" and "Failed" ...etc. into German before we displayed them, how would this effect this code ?

This won't affect the code as long as translations are applied only at display time (i.e., in the UI as the current code). The URL shortening logic relies on specific internal values, not translated ones. 

However, if translated values are used as actual params in the URL, then `minifyState()` won't recognize them. As a result:
 - The values won't be minified before compression.
 - The compressed URL will be longer.
 - But functionality wil still work, the decoded state will still contain the correct translated values.
 
  Galasa services may not interpret them correctly, since the backend expects known internal values like "Passed" and not "Bestanden". So, we always keep internal state and query values in English 🙂.

techcobweb (approved):

PR #161, opened 2025-07-16
Graphical Representation of Test Runs Table (#2328)

### LINKED ISSUE #2328: Web UI: Graphical representation of test runs table
## Story
As a Galasa user looking at test results, I want to view the results in a graphical view rather than in a table view.

## Background

Image

## Tasks
- [ ] New tab in the test runs page
- [ ] Spinney thing as it loads the results
- [ ] Time Axis with 'time' label
- [ ] Page only built when the user tries to view the graph
- [ ] 45 degree text labels showing the date/time to second granularity
- [ ] Closely-packed labels on the axis can be skipped so it looks cleaner
- [ ] If labels start overlapping, we may need a horizontal scroll bar underneath
- [ ] Each test run represented with a dot using the status color (Passed: green, Failed: red, other: amber...)
- [ ] Each dot has hover-help which shows the information in the table design
- [ ] Click on a dot and it navigates to the test run page for that dot
- [ ] URL contains the `&tab=graph` so we can navigate back to the exact same place

shivambisen:
##  Description

This PR implements a new **Graph** tab for the Test Runs page, allowing users to view their test results in a scatter‑plot format instead of—or alongside—the traditional table view.

**Related Issues:**  
closes [#2328](https://github.com/galasa-dev/projectmanagement/issues/2328)

  - Added “Experimental Features” checkbox to toggle Graph tab.
  - Updated Test Runs tab to include a Graph tab (hidden by default).
  - Created `TestRunGraph` React component
  - Hooked click events on SVG dots to route to `/test-runs/{id}` with breadcrumb push.
  - Adjusted chart container styles in `TestRunsGraph.module.css`.

---

## Demo 

https://github.com/user-attachments/assets/e7e3048b-19aa-4331-b947-fdbb6396feee


Tasneemmohammed0:
Thank you for raising this PR, @shivambisen! I ran it locally and it looks really good to me. Great job! 
Also, I really liked the idea of filtering by status in the Graphs tab. Very nice touch! 🤩

I just have a few small thoughts and suggestions. Feel free to start a Slack thread so the team can discuss further.

1. The hover tooltip works fine when the graph isn’t too crowded, but with denser data, there's some flickering where the label seems to render twice.

2. Not sure if this is a Carbon Charts issue, but when the selected date range is only one day, the dots appear at the far corners of the chart, as shown here:

image

3. Would it be better for the displayed date to be the actual submitted date? It looks inconsistent to me right now. I think it would be better if the rendered date matched where the dot actually appears, so we can get a clearer sense of the test run’s submitted date without needing to hover.

image

4. The graph sometimes doesn’t behave well when it’s dense, like in this photo. Maybe a horizontal scroll bar at the bottom would help? Not sure if that’s possible in this custom component. I’m thinking: should we maybe discuss a new feature with the team to zoom in/out on the graph, would that be better?

image

5. Really minor but I think the `EnvFail` dot could be another color besides blue, to better reflect that it's actually a failure. Maybe orange would be good?

Tasneemmohammed0:
I also think it would look nicer to show a chart skeleton, like the one in the results table, instead of a loading spinner. 
You can check out the skeleton example here: https://charts.carbondesignsystem.com/scatter

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,176 @@
...
+      `<strong>${header}:</strong> ${run[key] ?? "Unknown"}`
+    ).join("<br/>")}
+            </div>`;
+      },
+    },
+    legend: { alignment: "center" },
+    data: { loading: isLoading },
+    toolbar: { enabled: false },
+    experimental: true,
+  }),[headerDefinitions, isLoading, isLightTheme]);
```
Tasneemmohammed0: `resultColorMap` and `translations` are missing dependencies here.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,176 @@
...
+    },
+    height: "400px",
+    points: { radius: 5, fillOpacity: 1 },
+    color: { scale: resultColorMap },
+    tooltip: {
+      enabled: true,
+      customHTML: (points: any[]) => {
+        const run = points[0]?.custom || {};
+        return `
+            <div style="padding:6px; font-size:0.9rem;">
```
Tasneemmohammed0: Consider extracting the `customHTML` function into a separate helper inside this file to improve readability and avoid inline styles.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  useEffect(() => {
+    const container = chartContainerRef.current;
+    if (!container) return;
+
+    const handleClick = (event: MouseEvent) => {
+      let element = event.target as HTMLElement | null;
+      while (element && !(element as any).__data__) {
+        element = element.parentElement;
+      }
```
Tasneemmohammed0, line 203: I understand that this raw DOM traversal is necessary because Carbon Charts doesn't expose an `onClick` handler. However, it would be helpful to add a brief comment above this `useEffect` explaining why we're manually attaching a click listener and how it's extracting the clicked data point.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <div className={styles.spinnerWrapper}>
+        <Loading withOverlay={false} description={translations("loadingGraph")} />
+      </div>
+    );
+  if (!runsList.length) return <p>{translations("noTestRunsFound")}</p>;
+
+  // Compute timeframe text
+  const earliestDate = new Date(
+    Math.min(...runsList.map((r) => new Date(r.submittedAt).getTime()))
+  );
```
Tasneemmohammed0: The `timeFrameText` is currently computed inside the `TestRunsTable.tsx` component. It would be cleaner to lift this computation up to `TestRunsTabs.tsx` and pass it down as a prop, or even better, extract it into a reusable helper in the `utils/` folder. That way, we avoid duplicating logic and keep the responsibility for time range formatting in one place.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,176 @@
...
+          from: earliestDate.toLocaleString().replace(",", ""),
+          to: latestDate.toLocaleString().replace(",", ""),
+        })}
+      </p>
+      <div ref={chartContainerRef}>
+        <ScatterChart data={chartData} options={chartOptions} />
+      </div>
+    </div>
+  );
+}
```
Tasneemmohammed0: It would be beneficial to have some unit tests for this component to cover key behaviors, such as graph rendering, loading states, error handling, and user interactions like clicking on data points. Please consider adding tests to ensure these scenarios are properly validated.

Tasneemmohammed0 (commented):
Overall, the implementation looks clean and well-structured! I've added a few minor suggestions regarding code refactoring and adding unit tests to cover the new functionality. 🙂

shivambisen:
@Tasneemmohammed0  I’ve made all the requested changes  thanks for the helpful suggestions!

Just had one question regarding your point:

> Would it be better for the displayed date to be the actual submitted date? It looks inconsistent to me right now. I think it would be better if the rendered date matched where the dot actually appears, so we can get a clearer sense of the test run’s submitted date without needing to hover.


That’s a great observation it would definitely help with clarity. Just to confirm, are you suggesting we display all dates along the axis for consistency, or only those dates where test runs exist, to keep the graph cleaner and more focused?

Tasneemmohammed0:
Thank you for making those changes, @shivambisen! 🤩 I ran it locally and it looks great, I really liked the Zoom In/Out functionality!
Could you please update the preview video so that @techcobweb and @eamansour can take a look at the final changes and share their feedback?

Tasneemmohammed0:
> @Tasneemmohammed0 I’ve made all the requested changes thanks for the helpful suggestions!
> 
> Just had one question regarding your point:
> 
> > Would it be better for the displayed date to be the actual submitted date? It looks inconsistent to me right now. I think it would be better if the rendered date matched where the dot actually appears, so we can get a clearer sense of the test run’s submitted date without needing to hover.
> 
> That’s a great observation it would definitely help with clarity. Just to confirm, are you suggesting we display all dates along the axis for consistency, or only those dates where test runs exist, to keep the graph cleaner and more focused?

Yep, I’d go with the second option, displaying dates only where test runs exist. That way, the graph stays clean and focused while still giving enough context. And when there are a lot of test runs in a certain area, we can skip some dates to reduce clutter, knowing that we can zoom in to see the exact dates if needed.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const resultColorMap: Record<string, string> = {
+    passed:    "#2ecc40", // bright green
+    failed:    "#ff4136", // vivid red
+    envfail:   "#ffb700", // gold/yellow
+    cancelled: "#ab47bc", // medium purple
+    requeued:  "#00bcd4", // cyan
+    "n/a":     "#bdbdbd", // light gray
+    other:     "#607d8b", // blue-gray
+  };
+
```
Tasneemmohammed0, line 78: `resultColorMap` is giving me a warning in VSCode. We can either wrap it in a `useMemo` hook or move it outside the component and remove it from the dependency array.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+"use client";
+
+import "@carbon/charts/styles.css";
+import { ScatterChart } from "@carbon/charts-react";
+import { ScaleTypes } from "@carbon/charts";
+import { Loading, InlineNotification } from "@carbon/react";
```
Tasneemmohammed0: `Loading` is an unused import now.

shivambisen:
Thanks, @Tasneemmohammed0 ! Here's the updated preview:

https://github.com/user-attachments/assets/6a60a6d1-6d2d-49b9-94dc-04b9d64aeb2b

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -100,13 +101,12 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
     const dates = runsList.map((run) =>
       new Date(run.submittedAt || 0).getTime(),
     );
-    const earliestDate = new Date(Math.min(...dates));
-    const latestDate = new Date(Math.max(...dates));
+    const { earliest, latest } = getEarliestAndLatestDates(runsList.map(r => r.submittedAt));
```
Tasneemmohammed0: We can pass the dates array here now, since it's currently unused.

galasa-ui/src/tests/components/TestRunGraph.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  it("shows no data message", () => {
+    render(<TestRunGraph {...defaultProps} runsList={[]} />);
+    expect(screen.getByText(/No test runs found/)).toBeInTheDocument();
+  });
+
+  it("shows limit exceeded warning", () => {
+    render(<TestRunGraph {...defaultProps} limitExceeded={true} />);
+    expect(screen.getByText(/Limit Exceeded/)).toBeInTheDocument();
+  });
```
Tasneemmohammed0, line 136: It would be good to add a unit test to verify navigation to the single test run page when a test run dot is clicked.

Tasneemmohammed0: Also, when no visible columns are selected, we should display a message, similar to the one shown in the results table. Once that’s added, we can also verify it with a unit test.

Tasneemmohammed0 (commented):
Just gave it another look. Everything’s coming together nicely! I’ve added a few more small comments, mostly minor stuff.

shivambisen:
Yes, its looking better now 

screencapture-localhost-3000-test-runs-2025-07-23-03_39_48

Tasneemmohammed0 (approved):

Tasneemmohammed0:
Following the meeting discussion, I received a few comments from the team. They also liked how it looks overall:
1. “Graphs” should be changed to “Graph”.
2. It's better to always show all columns in the hover tooltip, regardless of their visibility in the graph. This prevents the graph from disappearing when all columns are hidden. The tooltip content should be independent from the visible columns, as there's enough space for it.
3. If there are thousands of test runs on a single day, how will the graph handle that? Will the data points become too dense or overlap?

**Side Note:**  The main branch now includes a new hook for date formatting. The graph dates should now use the `formatDate` function from that hook for consistency.

shivambisen:
> If there are thousands of test runs on a single day, how will the graph handle that? Will the data points become too dense or overlap?
```
const totalRuns = runsList.length;
const pointRadius = totalRuns > 500 ? 2 : totalRuns > 200 ? 3 : 4;
```
This approach helps reduce visual clutter by making points smaller as the data volume increases, While it doesn’t eliminate overlap entirely if points are in huge number, it does improve clarity significantly when handling dense datasets.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const searchParams = useSearchParams();
+  const { pushBreadCrumb } = useHistoryBreadCrumbs();
+
+  const chartContainerRef = useRef<HTMLDivElement>(null);
+
+  const headerDefinitions = useMemo(() => {
+    if (!runsList.length) return [];
+    const firstRun = runsList[0];
+    return Object.keys(firstRun)
+      .filter((key) => key !== "submittedAt" && key !== "id") // exclude keys you don't want in tooltip
```
Tasneemmohammed0: It would be better to include `submittedAt` in the tooltip.

Tasneemmohammed0 (commented):
We’re getting really close, great progress so far!
One last thing I noticed: when the selected date range is only a single day, the data points still appear at the far edges of the chart.

image

image

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+  isError?: boolean;
+}
+const resultColorMap: Record<string, string> = {
+  passed:    "#2ecc40", // bright green
+  failed:    "#ff4136", // vivid red
+  envfail:   "#ffb700", // gold/yellow
+  cancelled: "#ab47bc", // medium purple
+  requeued:  "#00bcd4", // cyan
+  "n/a":     "#bdbdbd", // light gray
+  other:     "#607d8b", // blue-gray
```
eamansour: Can these values be defined as constants? That way, we wouldn't need to add comments on what the values represent. We already have a `COLORS` constant defined in `src/utils/constants/common.ts`, so feel free to add and import colours from there.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const handleClick = (event: MouseEvent) => {
+      let element = event.target as HTMLElement | null;
+      while (element && !(element as any).__data__) {
+        element = element.parentElement;
+      }
+      const dataPoint = element && (element as any).__data__;
+      if (!dataPoint?.custom) return;
+
+      const selectedRun = dataPoint.custom as runStructure;
+      if (!selectedRun.id) return;
```
eamansour: Can we avoid having multiple returns in this function? Having multiple returns can make the code more challenging to follow and come back to.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+        ticks: {
+          values: xTickValues,
+          formatter: (tick: number | Date, i?: number) => {
+            const date = tick instanceof Date ? tick : new Date(tick);
+            const day = date.getDate();
+            const month = date.toLocaleString("default", { month: "short" });
+            if (day === 30) {
+              return `${month} ${day}`;
+            }
+            return `${day}`;
```
eamansour: Would be good to store the value to return in a variable and then have one return statement at the end here.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+        title: translations("submittedAt"),
+        mapsTo: "date",
+        scaleType: ScaleTypes.TIME,
+        ticks: {
+          values: xTickValues,
+          formatter: (tick: number | Date, i?: number) => {
+            const date = tick instanceof Date ? tick : new Date(tick);
+            const day = date.getDate();
+            const month = date.toLocaleString("default", { month: "short" });
+            if (day === 30) {
```
eamansour: It's not immediately obvious why `day == 30` is significant. What about months with 31 days? Or February, when there's 28/29 days?

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    container.addEventListener("click", handleClick);
+    return () => container.removeEventListener("click", handleClick);
+  }, [chartData, pushBreadCrumb, router, searchParams]);
+
+  if (isError) return <p>{translations("errorLoadingGraph")}</p>;
+  if (isLoading)
+    return (
+      <div className={styles.spinnerWrapper}>
+        <SkeletonText width="100%" style={{ height: 400 }} />
```
eamansour: Please avoid using inline styling. This mixes CSS code with HTML code. It would be good to maintain a separation of these things.

Can we give this skeleton a CSS class? Or import one that already exists?

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+      </div>
+    );
+  if (!runsList.length) return <p>{translations("noTestRunsFound")}</p>;
```
eamansour: There's a bit of inconsistency in the formatting of `if` statements. It would be good to use the format:
```ts
if (<condition>) {
  <code>
}
```

so that the codebase remains consistent throughout.

galasa-ui/src/components/test-runs/TooltipTestRunGraph.tsx
```diff
@@ -0,0 +1,14 @@
...
+ */
+export function getTooltipHTML(points: any[], headerDefinitions: { key: string, header: string }[], styles: any) {
+  const run = points[0]?.custom || {};
+  return `
+    <div class="${styles.tooltipContent}">
+      ${headerDefinitions.map(({ key, header }) =>
+    `<strong>${header}:</strong> ${run[key] ?? "Unknown"}`
+  ).join("<br/>")}
+    </div>`;
+}
```
eamansour: Since this is in the `components` directory, I would expect this to return a JSX element like all the other components. Components should be things that you can render like `<MyComponent ... />`.

This looks like a helper function that returns a string of HTML.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+      const dateKey = date.toISOString().split("T")[0];
+      const count = (dateMap[dateKey] || 0) + 1;
+      dateMap[dateKey] = count;
+
+      return {
+        group: (run.result || "other").toLowerCase(),
+        date,
+        value: count,
+        custom: run,
+      };
```
eamansour: Might be better for readability to move this map function out into a separate function and then call that inside the `runsList.map` call - like:

```ts
runsList.map((run) => yourFunction(run))
```

galasa-ui/src/components/test-runs/TestRunGraph.tsx
```diff
@@ -0,0 +1,205 @@
...
+  // Therefore, we manually attach a click event listener to the chart container.
+  // The handler traverses the DOM from the event target upwards to find an element with a __data__ property,
+  // which contains the chart's data point object. This allows us to extract the clicked run and perform navigation.
+  useEffect(() => {
+    const container = chartContainerRef.current;
+    if (!container) return;
+
+    const handleClick = (event: MouseEvent) => {
+      let element = event.target as HTMLElement | null;
+      while (element && !(element as any).__data__) {
```
eamansour: Using `any` defeats the purpose of using TypeScript and should really be used sparingly. Do we know what `element` should be here?

Same applies throughout this PR where `any` is used.

eamansour (commented):
Looking good! Thanks for making these changes - I've added a few thoughts and suggestions, feel free to let me know if anything doesn't make sense!

eamansour:
> > If there are thousands of test runs on a single day, how will the graph handle that? Will the data points become too dense or overlap?
> 
> ```
> const totalRuns = runsList.length;
> const pointRadius = totalRuns > 500 ? 2 : totalRuns > 200 ? 3 : 4;
> ```
> 
> This approach helps reduce visual clutter by making points smaller as the data volume increases, While it doesn’t eliminate overlap entirely if points are in huge number, it does improve clarity significantly when handling dense datasets.

I can't seem to find where that code is in the PR - we could modify it to be more dynamic rather than hard-coding numbers into it. For example, we could start with a maximum number of runs (e.g. 500), then update the point radius based on the percentage of runs we're currently displaying (within a certain range - like between 1 and 5 for example).

Side note: Using nested ternary statements for the `pointRadius` makes this code more complex than it needs to be! It would be great to always try to make code as easy to understand as possible to make maintaining the code simple for other developers, even if you need to write more lines of code to do so.

galasa-ui/src/components/test-runs/TestRunGraph.tsx
shivambisen: You're absolutely right  the condition day === 30 was a workaround from an earlier implemntation. Initially, I wasn’t sure how to remove the default highlight on the 30th, so I temporarily used it to attach the month label as a reference point(as it become easy to see months).

However, in the latest changes, I’ve refactored the logic so that the month is now correctly attached to the 1st of every month, and the special highlighting for the 30th has been removed.

shivambisen:
@eamansour Sorry it took a bit longer than expected. 
I've made all the required updates
Please Let me know your thoughts

techcobweb:
Looks like the unit tests are failing ? Is that fixable ? Seems this one is so close to being good it would be a shame to not get it over the line.

@James-Cocker is an optional reviewed. If we're good to go then we can merge this one ?

James-Cocker:
Looks great @shivambisen! I love the UI and how the graphs fit together - especially how smooth and intuitive the timescale feature is with the graph. Looks like it's just the tests in TestRunsTabs.test.tsx that are failing, can I help at all?

Edit: Apologies, just saw your commit so I'll take another quick look now.

shivambisen:
@techcobweb @James-Cocker  Apologies for catching this a bit late . I should have checked the build earlier. The test was failing due to a minor mismatch:
submittedAt: "N/A" was actually coming through as submittedAt: "-".

James-Cocker:
Don't be silly there's no need to apologise, we're all very excited for this change! 

Just wanted to mention a couple things i saw after maxing out the duration at 90 days. 

1) It seems like the dates overlap, I'm wondering if there's a way to remove them or make them smaller if they start to do this (see below)
image

2) Any chance we could get the cursor to be of type pointer when hovering over a test run? Quite minor but I think it'd be a lot more intuitive 😄.

Anyways it looks amazing, great stuff.

James-Cocker (approved):

James-Cocker:
Mike and I were thinking we should get these changes through now, then raise my previous point as a defect later down the line.

techcobweb:
@shivambisen Would like to merge this feature branch a.s.a.p. but one of the commits on it isn't signed. So we need you to fix that. Can I suggest you squash all your commits together into one and sign that ? or else go back and fix your commits, then force-push your branch ? 

I could squash it as I merge it in, but that would wipe out any of your signatures and wouldn't show you as the author in the git log... which I would rather avoid.

shivambisen:
I've cleaned things up it's not exactly a squash, but I did run into some challenges while fixing signatures on merge commits.
The good news is I finally figured out how to properly handle signed merge commits, so things should be in order now

techcobweb (approved):

PR #162, opened 2025-07-17
Enable download of all artifacts and logs as a zip folder

### LINKED ISSUE #2302: Web UI: Download everything to do with a single test run.
## Story
As a tester I want to be able to download a zip file containing all of the artifacts and run logs to my local file system.

## Background
From the web UI, you can download individual artifacts, but can't download the whole lot of information about a test run.

## Tasks
- [x] Add a download icon to the title bar in the test run page
- [x] Hover help something like "download all artifacts for this test run"
- [x] It creates a zip with all of the artifacts inside, and the user can save it where they like.

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2302](https://github.com/galasa-dev/projectmanagement/issues/2302)
## Changes
- Added a download icon to the title bar of the Test Run page.
- On click, all artifacts and the run log are fetched, zipped, and offered as a single `.zip` file for download.
- If the download fails, a clear error message is shown.

> [!Note]
> *Edited:* Zipping logic is now handled on the server side to improve performance when downloading large artifacts.


techcobweb:
If the run.log is massive, will it still download ?

jadecarino (approved):
Looks good to me, Tasneem! I am not a front-end expert but I can't see anything obviously wrong and glad to see a good unit test and comments in here, so was readable/understandable for me. 

Approved, assuming you will respond to Mike's comment above - do you know if there is a certain size of data that would eventually cause this to fail? Is this maybe a good additional unit test? I'll let you decide/discuss with Eamonn on Monday.

Tasneemmohammed0:
> If the run.log is massive, will it still download ?

Currently, the zipping is handled on the client side, which relies on the browser's memory, and that can be quite limited. For now, this works fine given the typical size of our artifacts and logs. However, if run.log or the total artifact size grows significantly in the future, it would make sense to move the zipping logic to the server side to ensure better stability.

galasa-ui/src/components/test-runs/TestRunDetails.tsx
```diff
@@ -112,12 +118,66 @@ const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsP
...
+              data-testid="icon-download-all"
+            >
+              {isDownloading ? 
+                <Loading 
+                  small
+                  withOverlay={false}
+                /> 
+                : 
+                <CloudDownload size={20}/>}
+            </button>
```
eamansour: Could we use the carbon design button and tooltip like we've done for the individual artifacts to maintain consistency?

Also the tooltip seems to be padded a lot more than the download button for the individual artifacts, and I don't think we need to capitalise "Artifacts" in "Download all artifacts".

So instead of:
Screenshot 2025-07-22 at 15 45 44


The button would visually behave like:
Screenshot 2025-07-22 at 15 45 09

galasa-ui/src/app/internal-api/test-runs/[runId]/zip/route.ts
```diff
@@ -0,0 +1,80 @@
...
+      headers: headers,
+    });
+  } catch (err) {
+    console.error(`Failed to create zip for run ${runId}:`, err);
+    return new Response(JSON.stringify({ error: 'Failed to generate the zip file on the server.' }), {
+      status: 500,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
```
Comment (eamansour) [inline galasa-ui/src/app/internal-api/test-runs/[runId]/zip/route.ts]: We could split parts of this GET function up into smaller functions to make it easier to read through?

galasa-ui/src/components/test-runs/TestRunDetails.tsx
eamansour: This would likely mean we need to do something similar for the share button to keep it consistent with the download button. Happy for this to be done in a separate PR.

eamansour (approved):
Looks good! Just a few small comments but approved assuming they'll be addressed

galasa-ui/src/components/test-runs/TestRunDetails.tsx
Tasneemmohammed0: I will open a new issue to implement the suggested enhancements for the share button 👍

eamansour (approved):

PR #163, opened 2025-07-17
fix(ui): ensure 'Next' button stays in viewport during log search

### LINKED ISSUE #2304: Web UI: Test run page: Run log tab: finding next occurrance scrolls the 'next' button off screen...
## Describe the bug
- Go to a test run
- Search for something like `ERROR`
- Go to the next occurrence using the 'next' button
- The contents of the run log pane scrolls to the next occurrance, but the 'next' button moves out of the visible window viewport, so it's hard to go to the next occurrance.

Tasneemmohammed0:
## Why?
Fixes [https://github.com/galasa-dev/projectmanagement/issues/2304](https://github.com/galasa-dev/projectmanagement/issues/2304)


galasa-ui/src/styles/LogTab.module.css
```diff
@@ -78,7 +78,7 @@
   border-radius: var(--border-radius);
   padding: 1rem;
   font-family: 'IBM Plex Mono', monospace;
-  height: 100vh;
+  height: 60vh;
```
eamansour: Instead of making the content height smaller, would it be better to make the log tab's search bar and other controls stick to the top of the screen when you scroll down past them (i.e. using `position: sticky` in the CSS)?

Then, when you scroll back up, the controls can go back to where they were originally.

eamansour (commented):
One suggestion - let me know if anything doesn't make sense!

eamansour (approved):

PR #164, opened 2025-07-21
Support user-configurable date and time formats

### LINKED ISSUE #2274: Web UI: Date and time formats should be in user preferences
## Story
As a Galasa user, I want to change the date and time formats I see when using the web UI, so that I can use my preferred formats without being forced into using a specific date and time format.

## Background
Users should have a preferences section in their settings page to allow them to change the date and time formats being used on various web UI pages. For example, some users may prefer DD/MM/YYYY and a 12hr time format, while others may want to use YYYY/MM/DD and a 24hr time format.

Once provided, all dates and timestamps received from the API server should be converted to the relevant date and time formats by the web UI.

Would it be beneficial for Galasa service admins to be able to set a default date and time format in the Helm values? Or should this be set to use the system default and then users can decide to change it themselves?

## Tasks
- [ ] ...

Tasneemmohammed0:
## Why?
Closes [#2274](https://github.com/galasa-dev/projectmanagement/issues/2274)
## Changes
- Added a new `DateTimeFormat` section to the settings page.
- Created a custom `useUserPreferences` hook to manage user-level settings.
- Introduced a React context `DateTimeFormatContext` for consistent formatting across the application.
- User-selected format is saved to local storage and loaded on startup.
- Refactored all existing date usages to rely on the centralized format provider.
- Added and adjusted unit tests to reflect the new logic.


techcobweb:
I am wondering whether local storage is the correct place for this really.

I think ideally the preference should be stored in the users' settings, and stored on the server over REST, but we don't have the REST service yet, so this seems sensible.

I suggest we raise a second story "Time format preferences to be stored on the server and shared across browser instances" or something. It would be dependent upon the implementation of this story: https://github.com/galasa-dev/projectmanagement/issues/2311

techcobweb:
Can you please do a small video capture of the feature being used, and/or some screen-shots ?
It's hard to get the look of it from the code without cloning your branch and building it locally...

Tasneemmohammed0:
> Can you please do a small video capture of the feature being used, and/or some screen-shots ? It's hard to get the look of it from the code without cloning your branch and building it locally...


https://github.com/user-attachments/assets/a1260659-6528-439d-a5b5-63e0ecfef8c6

eamansour:
> > Can you please do a small video capture of the feature being used, and/or some screen-shots ? It's hard to get the look of it from the code without cloning your branch and building it locally...
> 
> date-time.format.mp4

Thanks for adding that video - a couple of issues I can see from it:
- When selecting the de-DE date format, the dropdown shows the format as DD.MM.YYYY but the dates actually render to DD.M.YYYY
- The 12-hour format selection seems to crop out the AM/PM part of the format, showing only "A" at the end

Tasneemmohammed0:
>     * The 12-hour format selection seems to crop out the AM/PM part of the format, showing only "A" at the end

@eamansour , The "A" was actually intended to be a shorthand preview based on common formatting conventions. However, I realized now that this isn't very intuitive. I've updated the format string in the dropdown to use a more user-friendly preview.

galasa-ui/src/components/mysettings/DateTimeFormatSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+              <Dropdown
+                helperText={translations("selectTimeFormat")}
+                label={translations("selectTimeFormat")}
+                id="custom-time-format-dropdown"
+                items={TIME_FORMATS}
+                itemToString={(item: TimeFormat) => (item ? `${translations(item.label)} ${item.format}` : '')}
+                selectedItem={TIME_FORMATS.find(item => item.label === preferences.timeFormat)}
+                onChange={(e: {selectedItem: TimeFormat}) => handleChange(PREFERENCE_KEYS.TIME_FORMAT, e.selectedItem?.label || TIME_FORMATS[0].label)}
+                size="lg"
+                disabled={preferences.dateTimeFormatType !== 'custom'}
```
eamansour: Very minor - this line is the same as line 59, we could perhaps store the result of `preferences.dateTimeFormatType !== 'custom'` in a const and use that instead of duplicating the comparison?

galasa-ui/src/contexts/DateTimeFormatContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const defaultPreferences = {
+    [PREFERENCE_KEYS.DATE_TIME_FORMAT_TYPE]: 'browser',
+    [PREFERENCE_KEYS.LOCALE]: 'en-US',
+    [PREFERENCE_KEYS.TIME_FORMAT]: '12-hour'
+  };
+
+  const [preferences, setPreferences] = useState<DateTimeFormatContextType['preferences']>(() => {
+    if (typeof window === 'undefined') {
+      // Return default state during SSR
+      return defaultPreferences;
```
eamansour: Very minor - we could change this to only have one return statement at the end of this function?

galasa-ui/src/utils/types/dateTimeFormat.tsx
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+
+export type DateTimeFormats = 'custom' | 'browser'
+export type Locale = { code: string; format: string; example: string };
+export type TimeFormat = { label: string; format: string };
```
eamansour: We have the following `AmPm` type in `interfaces/common.ts`, would it be better off in this file now?

```ts
export type AmPm = 'AM' | 'PM';
```

eamansour (approved):

eamansour:
@Tasneemmohammed0 one more comment after trying the changes out locally:

When going to the test runs table, the "Submitted At" time doesn't seem to be rendering:
Screenshot 2025-07-22 at 16 35 24

Then in the console logs, I can see this error popping up:

Screenshot 2025-07-22 at 16 34 46

Tasneemmohammed0:
> @Tasneemmohammed0 one more comment after trying the changes out locally:
> 
> When going to the test runs table, the "Submitted At" time doesn't seem to be rendering: 

@eamansour I've pushed a fix for the browser incompatibility issue and addressed the required changes. Please verify on your end and confirm if everything works as expected before merging.

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -184,10 +185,10 @@ export default function TestRunsTabs({ requestorNamesPromise, resultsNamesPromis
   
     return runs.map((run) => {
       const structure = run.testStructure || {};
-  
+      console.log(structure.queued);
```
eamansour: I think we can remove this console.log line

eamansour (approved):
Thanks for making those changes, I tried running the new version out locally and the dates and times render correctly on the test runs table. One tiny cleanup comment to remove a console.log line but otherwise looks good!

PR #165, opened 2025-07-22
#2219 User Role should be on the settings panel rather than the profile page

James-Cocker:
# why?
https://github.com/galasa-dev/projectmanagement/issues/2219
- Added new role field in settings page, removed it from profile page.


galasa-ui/src/components/users/UserRole.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const roleDescription = userProfile.synthetic?.role?.metadata?.description;
+
+  return (
+    <div className={styles.profileDetails}>
+      { isLoading ?
+        <Loading data-testid="loader" small={false} active={isLoading} />
+        : !isError &&
+        <>
+          <div className={styles.userDetailsContainer}>
+            <h3>{translations("title")}</h3>
```
techcobweb: "title" seems a little generic. Think we should at least say what panel/section this label is for ?

techcobweb (approved):
One minor comment but looks good as far as I can tell.

galasa-ui/src/components/users/UserRole.tsx
James-Cocker: Renamed to "user-role-title"

galasa-ui/messages/en.json
```diff
@@ -118,12 +118,17 @@
...
     "noLastLogin": "No last login date",
     "lastLoginWeb": "Last logged in to this web application (UTC): {date}",
     "lastLoginToken": "Last logged in using a Galasa personal access token (UTC): {date}"
   },
+  "ProfileRole": {
+    "user-role-title": "User Role",
+    "change-role" : "User roles dictate what you can or can't do. Please contact your local Galasa administrator if you need this changed.",
+    "current-role": "You currently have the role",
+    "roleDescription": "Details about role"
+  },
```
Tasneemmohammed0, line 131: Really minor: could the keys here be adapted to camelCase for consistency?

Tasneemmohammed0 (changes requested):

galasa-ui/messages/en.json
James-Cocker: Of course! Apologies, I should've checked.

PR #166, opened 2025-07-23
Update README to include local webui dev without galasactl

James-Cocker:
# Why?
As brought up in the retrospective, not everyone will want to get galasactl up and running just to run the webui, hence this change will allow that.


README.md
```diff
@@ -15,6 +15,10 @@ This is the home of the web user-interface for Galasa.
 ## To build locally
 - Run the `build-locally.sh` script
 
+#### To build locally without galasactl
+- Replace the line in build-locally.sh from `gradle --warning-mode all --info --debug generateTypeScriptClient` to `gradle --warning-mode all --info --debug -PsourceMaven="https://development.galasa.dev/main/maven-repo/obr" generateTypeScriptClient`
+- Ensure you do not commit this change as a part of your branch if you choose to do this.
+
```
eamansour: Instead of adding an instruction for developers to manually get the build working without having built the galasa repo locally, we could just update the build-locally script to have a `SOURCE_MAVEN` environment variable that defaults to the development.galasa.dev URL if it's not set, and then we can pass the `SOURCE_MAVEN` value into the Gradle command like:

```
-PsourceMaven="${SOURCE_MAVEN}"
```

We already do this in other build-locally scripts - see https://github.com/galasa-dev/galasa/blob/main/modules/framework/build-locally.sh for an example.

James-Cocker: Hey Eamonn, thank you very much for the feedback! I've implemented your suggestions, does it look okay? I tested out the webui was building and running locally okay with the extra env this now requires, but it'd be great if you could double check this just in case, thanks!

README.md
```diff
@@ -13,6 +13,19 @@ This is the home of the web user-interface for Galasa.
...
 ## To build locally
+
+To build the code, it needs to get the REST interface definition from somewhere, to generate stubs in order to make calling the REST API easy.
+Hence, we need to tell the build where to get this definition from. 
+
+If you don't have the `galasa` repository built locally, then the default behaviour is to refer to a hosted version.
+
+If you do have a built `galasa` repository already, then we can refer to your `~/.m2` repository using the `SOURCE_MAVEN` environment variable.
+```shell
+export SOURCE_MAVEN="file://${HOME}/.m2/repository"
```
techcobweb, line 24: Did this work ? I think you said you were going to check ?

James-Cocker: Yes I checked, it worked for me however I thought it'd be best to ask Eamonn to double check it as well.

README.md
eamansour: Thanks for making those changes, James! They work for me, so I'm happy to approve this now!

README.md
eamansour: Works for me, so happy to approve this PR

eamansour (approved):

PR #167, opened 2025-07-24
Navigate to Log Line from Method Row in Test Run View

### LINKED ISSUE #2282: Web UI: Test run page, methods tab, icon to click for each method which jumps you to the first line of the run log which relates to that method.
## Story
As a Galasa user diagnosing a test run problem, I want to navigate from a method, to the location in the run log where that method exists.

## Tasks
- [ ] Click on the method, and it should take you to the run log, to the line that method relates to.

Tasneemmohammed0:
## Why?
Closes [#2282](https://github.com/galasa-dev/projectmanagement/issues/2282)
## Changes
- Clicking a method row in the methods table now redirects to the log tab and scrolls to the log line where the method starts.
- The URL is synchronized with the current active tab. When a method is clicked, it appends a line parameter `(&tab=log&line=54)`, allowing users to share direct links to specific log lines.
- Unit tests are added to cover the entire feature.


galasa-ui/src/components/test-runs/MethodsTab.tsx
```diff
@@ -133,7 +140,11 @@ function MethodsTab({ methods }: { methods: TestMethod[] }) {
               <TableBody>
                 {rows.map((row) => {
                   return (
-                    <TableRow key={row.id} {...getRowProps({ row })}>
+                    <TableRow 
+                      key={row.id} 
+                      onClick={() => onMethodClick(methodDetails[parseInt(row.id)])} 
+                      {...getRowProps({ row })}
+                    >
```
eamansour: Should we have some kind of indicator that clicking the row will take you to the method's starting line in the run log tab?

At the moment, it's difficult to tell that the row is clickable since the cursor doesn't change to a finger-pointing icon that you tend to see on clickable elements.

Would changing the cursor icon be enough? Happy for you to experiment!

eamansour (commented):
Looks good overall! Just one comment about making it clearer that clicking the method in the methods table takes you to the method in the run log.

galasa-ui/src/components/test-runs/MethodsTab.tsx
Tasneemmohammed0: I'm thinking if it would be good to have an icon with hover help showing "Go to the method line." But at that point, the user would need to click only on that icon to navigate, and not the whole row. If we did this, it would also need to be done in the results table, so maybe it's worth a separate story.
Changing the cursor icon for now sounds good.

eamansour (approved):

PR #168, opened 2025-07-24
refactor: Display "-" instead of "Invalid Date" and added additional null/invalid checks

### LINKED ISSUE #2350: Web UI: Invalid Date message in test run details when there is no date.
## Describe the bug
When I look at a test run overview, if any of the dates are null, they appear as "invalid date" text. Could we just display a "-" or something that doesn't make it look like Galasa set invalid data somewhere, rather than just not having a value for that field ?

eg:

Image

## Expected behavior
If the date isn't set, display "-" instead of "Invalid date"

Tasneemmohammed0:
## Why?
Fixes [2350](https://github.com/galasa-dev/projectmanagement/issues/2350)

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunDetails.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunDetails.tsx
@@ -69,10 +69,10 @@ const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsP
       package: runDetails.testStructure?.testName?.substring(0, runDetails.testStructure?.testName.lastIndexOf('.')) || 'N/A',
       requestor: runDetails.testStructure?.requestor!,
       rawSubmittedAt: runDetails.testStructure?.queued,
-      submitted: runDetails.testStructure?.queued ? formatDate(new Date(runDetails.testStructure?.queued!)) : 'N/A',
-      startedAt: runDetails.testStructure?.startTime ? formatDate(new Date(runDetails.testStructure?.startTime!)) : 'N/A',
-      finishedAt: runDetails.testStructure?.endTime ? formatDate(new Date(runDetails.testStructure?.endTime)) : 'N/A',
-      duration: (runDetails.testStructure?.startTime && runDetails.testStructure?.endTime) ? getIsoTimeDifference(runDetails.testStructure?.startTime, runDetails.testStructure?.endTime) : 'N/A',
+      submitted: runDetails.testStructure?.queued ? formatDate(new Date(runDetails.testStructure?.queued!)) : '-',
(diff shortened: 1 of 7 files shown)


galasa-ui/src/components/test-runs/TestRunDetails.tsx
```diff
@@ -69,10 +69,10 @@ const TestRunDetails = ({ runId, runDetailsPromise, runLogPromise, runArtifactsP
...
       requestor: runDetails.testStructure?.requestor!,
       rawSubmittedAt: runDetails.testStructure?.queued,
-      submitted: formatDate(new Date(runDetails.testStructure?.queued!)),
-      startedAt: formatDate(new Date(runDetails.testStructure?.startTime!)),
-      finishedAt: formatDate(new Date(runDetails.testStructure?.endTime!)),
-      duration: getIsoTimeDifference(runDetails.testStructure?.startTime!, runDetails.testStructure?.endTime!),
+      submitted: runDetails.testStructure?.queued ? formatDate(new Date(runDetails.testStructure?.queued!)) : 'N/A',
+      startedAt: runDetails.testStructure?.startTime ? formatDate(new Date(runDetails.testStructure?.startTime!)) : 'N/A',
+      finishedAt: runDetails.testStructure?.endTime ? formatDate(new Date(runDetails.testStructure?.endTime)) : 'N/A',
+      duration: (runDetails.testStructure?.startTime && runDetails.testStructure?.endTime) ? getIsoTimeDifference(runDetails.testStructure?.startTime, runDetails.testStructure?.endTime) : 'N/A',
```
eamansour: Should this be "-" instead of "N/A" like the expected behaviour in the issue mentions? Would that look better than "N/A"?

Same applies throughout this PR

eamansour (commented):
Changes look good, just one comment around what the placeholder should be when a date isn't set

galasa-ui/src/tests/contexts/DateTimeFormatContext.test.tsx
```diff
@@ -192,4 +192,16 @@ describe('DateTimeFormatContext', () => {
...
+  test('returns an empty text on invalid date', () => {
+    const invalidDate = new Date('invalid-date-string');
+    
+    render(
+      <DateTimeFormatProvider>
+        <TestComponent date={invalidDate} />
+      </DateTimeFormatProvider>
+    );
+
+    expect(screen.getByText(/Formatted Date:/)).toHaveTextContent('Formatted Date:');
```
eamansour: This assertion seems to be asserting what we've already retrieved? Should we check that the content has the placeholder we've set (e.g. "-")?

eamansour (approved):

PR #169, opened 2025-07-28
Update the Date Picker in test runs to align with the current date format preference

Tasneemmohammed0:
# Why ?
As part of [Web UI: Date and time formats should be in user preferences
#2274](https://github.com/galasa-dev/projectmanagement/issues/2274)

Make the date picker in the UI use the preferred date format.


galasa-ui/src/components/test-runs/DateTimePicker.tsx
```diff
@@ -22,6 +23,22 @@ interface DateTimePickerProps {
   onAmPmChange: (amPm: string) => void;
 }
 
+// Helper function to convert the display format to the Flatpickr format.
+const convertDisplayFormatToFlatpickr = (displayFormat: string): string => {
```
techcobweb: I think it's not clear what the input and output strings of this function is, or needs to be. Some examples in comments might help ? 

Some unit tests specifically for this would help also ?

galasa-ui/src/components/test-runs/DateTimePicker.tsx
```diff
@@ -53,15 +71,32 @@ export default function DateTimePicker({
...
 
+  const fullLocale = preferences?.locale || 'en-US';
+
+  // Look up the required Flatpickr format
+  const datePickerFormat = LOCALE_TO_FLATPICKR_FORMAT_MAP[fullLocale] || 'm/d/Y';
+  
+  // The `locale` prop in DatePicker expects a short language code (e.g., 'en', 'fr') for calendar translation.
+  const languageCodeForPicker = fullLocale.split('-')[0];
+
+  // Generate a placeholder from the Flatpickr format string.
```
techcobweb, line 66: What is a place holder ? The default current day/time ? `placeHolder` seems like it doesn't mean much.

galasa-ui/src/components/test-runs/DateTimePicker.tsx
```diff
@@ -22,6 +23,22 @@ interface DateTimePickerProps {
...
+    .replace(/MM/g, 'm')
+    .replace(/DD/g, 'd')
+    .replace(/YYYY/g, 'Y');
+};
+
+// Dynamically generate the map from SUPPORTED_LOCALES.
+const LOCALE_TO_FLATPICKR_FORMAT_MAP = Object.fromEntries(
+  SUPPORTED_LOCALES.map(({ code, format }) => [
+    code,
+    convertDisplayFormatToFlatpickr(format)
```
techcobweb: Will this work for all the locales ?

galasa-ui/src/components/test-runs/DateTimePicker.tsx
```diff
@@ -53,15 +71,32 @@ export default function DateTimePicker({
...
+        locale={languageCodeForPicker}
+        dateFormat={datePickerFormat}
         datePickerType="single"
         value={date}
         maxDate={new Date()}
         onChange={(dates: Date[]) => onDateChange(dates?.[0] || null)}
       >
-        <DatePickerInput id={`${legend}-date-picker`} labelText={translations('date')} placeholder="mm/dd/yyyy" />
+        <DatePickerInput
+          id={`${legend}-date-picker`}
```
techcobweb, line 80: Are we sure that {legend} won't contain a space or something that forms an invalid id ?
Are spaces allowed in ID's ? Is this not something we need to worry about ? Not sure.

techcobweb (approved):
Minor comments and questions. Approving anyway as it looks good, and you can decide how each thing is resolved ?

galasa-ui/src/components/test-runs/DateTimePicker.tsx
Tasneemmohammed0: Yes, we're sure. The legend is a prop, and it's always a single word.

galasa-ui/src/components/test-runs/DateTimePicker.tsx
Tasneemmohammed0: It will work for all locales that are currently specified in our program. But if we expand to support more, we'll need to hard-code them. Would using a hardcoded map be better for handling locale-specific formats like `dd-MMM-y`?

galasa-ui/src/components/test-runs/DateTimePicker.tsx
Tasneemmohammed0: The `placeHolder` is good when the user wants to type the date instead of picking it, so they know the format to type it in.

PR #171, opened 2025-07-29
James c mouse clickable on hover test runs table

James-Cocker:
# Why?
Seems much more intuitive that test entries could be clicked on with the cursor set to pointer on-hover. Unfortunately, inline styles had to be used as I exhausted every other option to apply the CSS.

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunsTable.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunsTable.tsx
@@ -162,7 +162,7 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
       );
     } else if (header === "submittedAt") {
       // Format the date using the context's formatDate function
-      cellComponent = <TableCell id="clickableItem">{formatDate(new Date(value))}</TableCell>;
+      cellComponent = <TableCell>{formatDate(new Date(value))}</TableCell>;
     }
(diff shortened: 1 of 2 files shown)


Tasneemmohammed0 (commented):
Good spot, James! We actually discussed that in Slack before, that we need something to point out that clicking on the test runs table actually takes you to the test run details. We thought about whether we need a specific icon or something, but anyway, this is probably worth another story to think about.
If we add it in the test runs table, we’d also add it to the methods table.
Maybe we should bring that up again. 🙂

Only one comment concerning inline styles: I tried to make the hover on the methods table by adding a classname, and it worked well that time. I think there shouldn’t be much difference between the two tables, did you try this?
https://github.com/galasa-dev/webui/blob/9f1ff38a1b8035a079da59335fd3038b6a6c8efa/galasa-ui/src/components/test-runs/MethodsTab.tsx#L146C22-L146C54

James-Cocker:
@Tasneemmohammed0 It worked! It seems like className="clickableRow" has much lower specificity level than className={styles.clickableRow}, which seems to be able to override the custom CSS on the TableRow element. Thanks for showing me!

Tasneemmohammed0 (approved):

techcobweb (approved):

PR #174, opened 2025-08-01
Implemented timeout for max record limit notification in test runs and incremented MAX_RECORDS

James-Cocker:
# Why?
Eamonn, Mike and I concluded this would provide better usability on the test-runs page, especially large datasets where each day could provide up to 1000 test results.


galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -64,11 +64,27 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
...
   })) || [];
 
+  // Set a timer to hide the notification after 5 seconds
+  useEffect(() => {
+    // Only show the notification if the limit was exceeded
+    if (limitExceeded) {
+      setShowNotification(true);
+      const timer = setTimeout(() => {
+        setShowNotification(false);
+      }, 6000); // 6 seconds
```
Tasneemmohammed0: Since we're standardizing all notification timeouts to 6 seconds, could we define `6000` as a constant and reuse it? Also, we should apply this constant in the latest changes on main as well.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -178,12 +194,14 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
...
-        className={styles.notification}
-        kind="warning" 
-        title="Limit Exceeded" 
-        subtitle={`Your query returned more than ${MAX_RECORDS} results. Showing the first ${MAX_RECORDS} records.`} 
-      />}
+      {limitExceeded && showNotification &&
+        <InlineNotification
+          className={styles.notification}
+          kind="warning" 
+          title="Limit Exceeded"
```
Tasneemmohammed0: This title should be translated as well.

Tasneemmohammed0 (commented):
Looks good to me 🙂  I just left two minor comments.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
James-Cocker: Done 👍

galasa-ui/src/components/test-runs/TestRunsTable.tsx
James-Cocker: Done 👍

PR #175, opened 2025-08-02
Add Timezone Preference Support with UI and Context Integration 

### LINKED ISSUE #2276: Web UI: Timestamps should be converted to the user's chosen timezone
## Story
As a Galasa user, I want timestamps shown on the web UI to appear in my preferred timezone, so that I can easily read and understand the timestamp without having to perform any manual conversions myself.

This is about the time zones, rather than the format of the time.

## Background
Timezones should be placed in user preferences, in a section called "Time Zone".

To contain a drop-down of all the timezones supported.

eg: 

Image

London, GMT

The backend Galasa service uses UTC for all of its timestamps. The web UI should convert timestamps received from the Galasa API server according to the timezone that a user has provided in their preferences.

Would it be beneficial for Galasa service admins to be able to set a default timezone in the Helm values? Or should this be set to UTC for everyone and then users can decide to change it themselves?

## Tasks
- [x] Timezone section in the settings page, with drop-down of options, plus a radio button with options like "Use the browser default timezone" "Use a specific timezone", with the greyed-out non-selected drop-down.
- [x] Store the selection in local storage
- [x] Settings -> Tokens. Each one has a time listed (created at). Change to "Created at 10/29/2024, 02:07:18 PM (GMT)" for example
- [x] Test run query needs to convert what times are collected into UTC based on this timezone
- [x] Test run table perma-URL should be in UTC
- [x] Test run table result "submitted at" field gets converted to preferred timezone.
- [x] Graph selected dot pop-up "submitted at" field gets converted to preferred timezone.
- [x] "Showing test runs submitted between 28/07/2025, 07:06:14 am and 28/07/2025, 10:32:32 am" --> "Showing test runs submitted between 28/07/2025, 07:06:14 am (GMT) and 28/07/2025, 10:32:32 am (GMT)"
- [x] User table fields "last login" and "last login token use" need to use this timezone.
- [x] user table titles change to "last login (GMT)" and "last login token use (GMT)" for example
- [x] edit other user-> Access tokens section needs to respect this also.
- [x] The profile page has this: which is confusing:
Image

Should be 
```
Last logged in to this web application: Mon, 28 Jul 2025 13:05:26 (GMT)
Last logged in using a Galasa personal access token: Thu, 19 Jun 2025 08:38:07 (GMT)
```
- [x] Unit tests for all the above
- [x] The detailed test run page->Overview tab has these inside:

Image

Which should change to :
```
Submitted: 29/07/2025, 07:02:44 am (GMT)
Started:   29/07/2025, 07:04:16 am (GMT)
Finished:  29/07/2025, 07:05:10 am (GMT)
```

Tasneemmohammed0:
## Why? 
Closes [2276](https://github.com/galasa-dev/projectmanagement/issues/2276)
## Changes
- Introduced a new Timezone Selection section in the user settings
- Persisted the user's selected timezone preference via `localStorage` (to be moved to server-side later)
- UI now displays dates in the user's preferred timezone
- Dates are formatted consistently using a centralized `DateTimeFormatContext`

> [!Note] 
For detailed implementation and usage guidelines, refer to `date-time-handling.md`


galasa-ui/docs/date-time-handling.md
```diff
@@ -0,0 +1,63 @@
...
+The formatDate function automatically handles everything:
+- It reads the user's chosen timezone (e.g., 'America/New_York') from the context.
+- It reads the user's chosen locale (e.g., 'en-GB') for correct date ordering (DD/MM/YYYY).
+- It reads the user's time format preference (12-hour or 24-hour).
+- It combines these to produce a correctly formatted, human-readable string.
+
+## UI-to-UTC Conversion Utilities
+In some cases, especially when a user is **inputting** a specific time, we need to perform the conversion in reverse: from UI-friendly parts into a single UTC Date object. Two utility functions exist for this purpose.
+
+#### `extractDateTimeForUI(date, timezone)`
```
techcobweb: These two functions might be better if they are called something like `dateTimeUTC2Local` and `dateTimeLocal2UTC` ?

galasa-ui/src/contexts/DateTimeFormatContext.tsx
```diff
@@ -78,32 +101,48 @@ export function DateTimeFormatProvider({ children }: { children: React.ReactNode
...
-        // Pass undefined to use the browser's default locale
-        formattedDate = new Intl.DateTimeFormat(undefined, options).format(date);
-      } else {
-        // Use the custom locale
-        formattedDate = new Intl.DateTimeFormat(locale, options).format(date);
-      }
+      // Define options specifically to extract the timezone name
+      const timeZoneNameOptions: Intl.DateTimeFormatOptions = {
+        timeZone: resolvedTimeZone,
+        timeZoneName: "short", // e.g., PST, EDT, EEST
```
techcobweb, line 121: What does the "short" mean ? should it be "utc" or something ?

Wondering now if someone just wants to set their preference to UTC un-adulterated dates. Is that something which is possible ?

galasa-ui/src/tests/components/UsersTable.test.tsx
```diff
@@ -43,19 +43,19 @@ jest.mock('next-intl', () => ({
...
+      "modalLabel": "Delete User",
+      "modalHeading": "Are you sure you want to delete '{user}'?",
+      "modalPrimaryButton": "Delete",
+      "modalSecondaryButton": "Cancel",
+      "modalNotificationTitle": "Deleting a user will remove any memory the Galasa service has of this user.",
+      "modalNotificationSubtitle": "Any access tokens previously allocated by this user will be removed and will no longer have access to the Galasa service.\n\nTests run by this user will remain untouched and available for queries.\n\nIf the user logs in to the Galasa service after this point, they will be challenged for credentials from the configured authorization authority, as they were originally when they started using the Galasa service. If that authority allows access, a new set of details for that user will be collected from that point.",
```
techcobweb, line 58: Seems a bit bad we have to re-define these to be exactly the same in the test bundle. Why do this ?

galasa-ui/src/contexts/DateTimeFormatContext.tsx
Tasneemmohammed0: The "short" here refers to the `timeZoneName` option in the `Intl.DateTimeFormat` API. It's used to format the timezone name in a short form. Like PST, UTC, EEST, etc.

If someone wants to see the raw, unadjusted UTC time, they can just select "UTC" from the timezone dropdown in the settings. That way, the `resolvedTimeZone` will be "UTC", and the formatted date will reflect UTC without any local offset.

I’ll record a short video to demonstrate this in the UI.

galasa-ui/docs/date-time-handling.md
```diff
@@ -0,0 +1,63 @@
+# Guide to Date and Timezone Handling in the Galasa UI
+This document outlines the standardized approach for handling all date and time-related operations within the codebase. The goal is to eliminate timezone ambiguity, ensure data consistency between the UI and server, and provide a clear and intuitive experience for users, regardless of their location.
```
techcobweb, line 2: Very nice set of notes. Clear and to the point. :-)

galasa-ui/docs/date-time-handling.md
techcobweb: just a suggestion really.

galasa-ui/src/contexts/DateTimeFormatContext.tsx
techcobweb: No need for a recording. I think I get it.

galasa-ui/docs/date-time-handling.md
Tasneemmohammed0: Thank you! :))

galasa-ui/src/tests/components/UsersTable.test.tsx
Tasneemmohammed0: We only need to mock the translations we’re checking for. They don’t have to match exactly, we can simplify or shorten them if that helps, though that would require updating some test checks.

techcobweb (approved):

PR #176, opened 2025-08-04
Enable Sharing of Specific Highlighted Lines in the Log Tab

Tasneemmohammed0:
## Why?
Refer to [2351](https://github.com/galasa-dev/projectmanagement/issues/2351)
## What's done?
- Added a Copy button to the search bar, enabled only when one or more lines are selected.
- Users can select specific lines in the log tab and click the Copy button to generate a permalink.
- Clicking the button appends the selected line range to the current URL.
- On initial mount, the URL is checked for selected lines, if found, the view scrolls to them and highlights them.
- Unit tested for selection and permalink generation.
- Unit tested for URL hashing handling.
## Demo (Updated)

https://github.com/user-attachments/assets/e85b11c9-10ea-41d8-b7b7-f0d2b545af93


galasa-ui/src/components/test-runs/LogTab.tsx
```diff
@@ -279,6 +330,40 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
...
+      const endNode = endPre.lastChild;
+      const endOffset = endNode.textContent?.length || 0;
+      range.setEnd(endNode, endOffset);
+
+      // Clear any existing selection and apply the new range
+      const selection = window.getSelection();
+      if (selection) {
+        selection.removeAllRanges();
+        selection.addRange(range);
+      }
```
eamansour: If a user wants to highlight a specific part of a line for whatever reason (e.g. copying a maven repository URL), this overrides that selection and will cause the entire line to be selected.

Is this the behaviour we want? It might be better to highlight entire lines by clicking and dragging down from the line numbers, allowing users to highlight individual parts of a line.

galasa-ui/messages/en.json
```diff
@@ -188,20 +188,21 @@
...
+    "matchCase": "Match case",
+    "matchWholeWord": "Match whole word",
+    "filtersMenuTitle": "Hide / Show Content",
+    "filterError": "Error",
+    "filterWarn": "Warning",
+    "filterInfo": "Info",
+    "filterDebug": "Debug",
+    "filterTrace": "Trace",
+    "downloadButton": "Download Run Log",
+    "copyPermalinkButton": "Copy Permalink"
```
eamansour: Instead of "Copy Permalink", the tooltip could say something like "Copy permalink with selected lines" or similar to make it clear that the permalink goes to the same run log but highlights the selected lines. I just think that "Copy Permalink" doesn't make it clear what the permalink is for.

Feel free to come up with a better tooltip!

galasa-ui/src/components/test-runs/LogTab.tsx
```diff
@@ -311,6 +401,75 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
...
+
+  // Effect to listen to the browser hash changes and updates the state that runs only on mount
+  useEffect(() => {
+    const handleHashChange = () => {
+      setCurrentHash(window.location.hash);
+    };
+
+    window.addEventListener('hashchange', handleHashChange);
+    return () => {
+      window.removeEventListener('hashchange', handleHashChange);
```
eamansour: Minor - we could store the event listeners in constants so we don't have to keep track of their occurrences if they ever change rather than string literals.

eamansour (commented):
Looks good! Just a few minor comments and a question around the intended highlighting behaviour

galasa-ui/src/components/test-runs/LogTab.tsx
Tasneemmohammed0: Resolved 👍

galasa-ui/src/components/test-runs/LogTab.tsx
```diff
@@ -514,9 +663,18 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
...
         />
+        <Button
+          kind="ghost"
+          renderIcon={Copy}
+          hasIconOnly
+          iconDescription={translations('copyPermalinkButton')}
+          onClick={handleCopyPermalink}
+          disabled={!selectedRange?.startLine}
+          data-testid="icon-button-copy-permalink"
+        />
```
eamansour: Just to make it clearer that this functionality exists, would it be worth having a tooltip that appears when the copy button is disabled to tell users to highlight lines in order to create a permalink with that selection visible?

At the moment, when the copy button is disabled, it's not clear what someone needs to do in order for it to become enabled!

eamansour (approved):
One last suggestion but otherwise looks good - approved!

galasa-ui/src/components/test-runs/LogTab.tsx
Tasneemmohammed0: Great Idea! Applied 👍

PR #182, opened 2025-08-05
James c 3270 feature flag

James-Cocker:
# Why?

Need a feature flag to be able to toggle on and off the 3270 panel. See https://github.com/galasa-dev/projectmanagement/issues/2322 for the full Epic. I also resolved a small build issue in the build-locally.sh file - this will be worth someone else double checking.


build-locally.sh
```diff
@@ -149,7 +149,7 @@ function generate_rest_client {
 
     # Execute Gradle command with SOURCE_MAVEN passed as a Gradle property
     gradle --warning-mode all --info --debug \
-      -PsourceMaven="$SOURCE_MAVEN" \
+      -PsourceMaven=SOURCE_MAVEN \
```
techcobweb: This change doesn't look good. The variable substitution which went before is probably important.

galasa-ui/messages/en.json
```diff
@@ -286,7 +287,7 @@
     "noTestRunsFound": "No test runs were found for the selected timeframe",
     "isloading":"Loading test results...",
     "limitExceededTitle": "Limit Exceeded",
-    "limitExceededSubtitle": "Your query returned more than {maxRecords} results. Showing the first {maxRecords} records.",
+    "limitExceededSubtitle": "Your query returned more than {maxRecords} results, please refine your search. Showing the first {maxRecords} records.",
```
techcobweb: We never say "please". Something like "Your query returned more than {maxRecords} results. Showing the first {maxRecords} records. To avoid this in the future narrow your time frame or change your search criteria to return fewer results."

Would that be better ?

galasa-ui/src/components/mysettings/ExperimentalFeaturesSection.tsx
```diff
@@ -18,13 +18,16 @@ export default function ExperimentalFeaturesSection() {
...
-      ,
+      label: translations(`features.testRunSearch`),
     },
     {
       key: FEATURE_FLAGS.INTERNATIONALIZATION,
       label: translations('features.internationalization'),
     },
+    {
+      key: FEATURE_FLAGS.THIRTY_TWO_SEVENTY_TERMINAL,
+      label: translations('features.thirtyTwoSeventyTerminal'),
```
techcobweb: Odd seeing 3270 spelt out, as everyone uses the numeric. Is features.3270 not allowed ?

galasa-ui/src/components/test-runs/TableDesignContent.tsx
```diff
@@ -64,6 +64,22 @@ export default function TableDesignContent({
     }
   };
 
+  const [showNotification, setShowNotification] = useState(selectedRowIds.length === 0);
+
+  // Set a timer to hide the notification after 5 seconds
```
techcobweb: "after 5 seconds" When the code changes this period value, this comment will get out of date. Suggest you comment "after a short period (seconds)" or something.

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -64,11 +64,27 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
...
   const [pageSize, setPageSize] = useState(10);
 
+  const [showNotification, setShowNotification] = useState(limitExceeded);
+  
   const headers = orderedHeaders?.filter(column => visibleColumns.includes(column.id)).map(column => ({
     key: column.id,
     header: translations(column.id)
   })) || [];
 
+  // Set a timer to hide the notification after 5 seconds
```
techcobweb: same as above

galasa-ui/src/components/test-runs/TestRunsTable.tsx
```diff
@@ -64,11 +64,27 @@ export default function TestRunsTable({runsList,limitExceeded, visibleColumns, o
...
+  useEffect(() => {
+    // Only show the notification if the limit was exceeded
+    if (limitExceeded) {
+      setShowNotification(true);
+      const timer = setTimeout(() => {
+        setShowNotification(false);
+      }, WARNING_NOTIFICATION_DURATION);
+
+      // Cleanup function: This will clear the timer unmounts before the timeout.
+      return () => clearTimeout(timer);
```
techcobweb: This code is duplicated. Might be worth defining a common function to show the notification and set the un-show up at the same time ?

We could then use the same thing for info notifications also (like "link has been copied to the clipboard")...

galasa-ui/src/tests/components/TestRunsTable.test.tsx
```diff
@@ -54,7 +54,8 @@ jest.mock("next-intl", () => ({
       "testName": "Test Name",
       "status": "Status",
       "result": "Result",
-      "pagination.of": "of {total}"
+      "pagination.of": "of {total}",
+      "limitExceededSubtitle": "Your query returned more than 2000 results, please refine your search. Showing the first 2000 records."
```
techcobweb: If 2000 is a constant you can get at, I would pull it from the code to make the testcase more resilient were it to ever change.

galasa-ui/src/utils/constants/common.ts
```diff
@@ -21,8 +21,11 @@ const COLORS = {
   BLUE_GRAY: "#607d8b"
 };
 
-// Maximum number of records to fetch in one go
-const MAX_RECORDS = 1000; 
+// Maximum number of records to fetch in one go.
```
techcobweb: These are test run records. We may have records of other things (user records, test case records). Seems that MAX_RECORDS isn't specific enough ?

galasa-ui/src/utils/constants/common.ts
```diff
@@ -21,8 +21,11 @@ const COLORS = {
...
+const MAX_RECORDS = 2000; 
+
+// 6 second duration for warning notifications.
+const WARNING_NOTIFICATION_DURATION = 6000;
```
techcobweb: I follow a general rule of including the time units in the constant name. So this would be easier to understand if it was `WARNING_NOTIFICATION_VISIBLE_MILLISECS` ?

This from having to spend ages working out what the time units are for a variable or constant when reading code.

The comment above just mirrors that, and repeating the value isn't needed... I would get rid of the comment. If the name is better, why have the comment ?

galasa-ui/src/utils/constants/common.ts
techcobweb: Perhaps `MAX_DISPLAYABLE_TEST_RUNS` or similar ?

Then the comment isn't really needed either.

galasa-ui/src/utils/featureFlags.ts
```diff
@@ -8,13 +8,15 @@
 export const FEATURE_FLAGS = {
   TEST_RUNS: 'testRuns',
   INTERNATIONALIZATION: 'internationalization',
+  THIRTY_TWO_SEVENTY_TERMINAL: 'thirtyTwoSeventyTerminal',
```
techcobweb: Maybe `IS_3270_SCREENS_ENABLED` or `IS_3270_SCREENS_DISPLAYABLE` is better? 
I assume it's holding a boolean, so the 'IS' hints at that.

techcobweb: Then you could use `is3270ScreensDisplayable` elsewhere also.

galasa-ui/messages/en.json
techcobweb: Why don't we say "please" ? I'm sure there is a good reason, but it's about being factual rather than pally. We could ask our documentation people for guidance I guess.

Error messages should always have 3 parts: 
- say what the problem is
- say what the system did about it for now
- say what the user needs to do in order to make things better

Most products I've worked on in the past add a 4th part:
- a unique identifier for the error, so people can relay that number to support channels, rather than the (possibly translated) error message.

I think in recent years this has gone out of favour for web UIs, but I'm open to the idea if we want to organise our user-visible error messages and give them all unique numbers. Or not. eg: `E34567`

Anything not a web UI (REST interface, CLI...etc) I would insist on it being numbered.

galasa-ui/src/components/mysettings/ExperimentalFeaturesSection.tsx
James-Cocker: Unfortunately not

build-locally.sh
James-Cocker: Fixed 👍

galasa-ui/messages/en.json
James-Cocker: Fixed 👍

galasa-ui/src/components/test-runs/TableDesignContent.tsx
James-Cocker: Fixed 👍

galasa-ui/src/components/test-runs/TestRunsTable.tsx
James-Cocker: Fixed 👍

galasa-ui/src/tests/components/TestRunsTable.test.tsx
James-Cocker: Fixed 👍

galasa-ui/src/utils/constants/common.ts
James-Cocker: Done 👍

galasa-ui/src/utils/constants/common.ts
James-Cocker: Done 👍

galasa-ui/src/utils/featureFlags.ts
James-Cocker: Done 👍

galasa-ui/src/components/test-runs/TestRunsTable.tsx
James-Cocker: Done 👍

galasa-ui/src/components/test-runs/TableDesignContent.tsx
```diff
@@ -13,8 +13,9 @@ import { useTranslations } from "next-intl";
 import { InlineNotification } from "@carbon/react";
 import { ColumnDefinition } from "@/utils/interfaces";
 import { sortOrderType } from "@/utils/types/common";
-import { DEFAULT_VISIBLE_COLUMNS, RESULTS_TABLE_COLUMNS } from "@/utils/constants/common";
-import { Dispatch, SetStateAction } from "react";
+import { DEFAULT_VISIBLE_COLUMNS, RESULTS_TABLE_COLUMNS, NOTIFICATION_VISIBLE_MILLISECS } from "@/utils/constants/common";
+import { Dispatch, SetStateAction, useEffect, useState } from "react";
+import { useNotification } from "@/components/common/UseNotification";
```
techcobweb: name could be more descriptive, about auto-disappearing...etc ?

techcobweb (approved):

galasa-ui/src/components/common/UseNotification.tsx
```diff
@@ -0,0 +1,21 @@
+import { NOTIFICATION_VISIBLE_MILLISECS } from '@/utils/constants/common';
+import { useEffect, useState } from 'react';
+
+export const useNotification = (initialState: boolean) => {
+  const [showNotification, setShowNotification] = useState(initialState);
+
```
Tasneemmohammed0: I really like the idea of extracting this into a hook, especially since we'll be using it quite a bit. Just a small suggestion: could you move the file to the hooks folder and rename it to `useNotification` (with a lowercase "u") to align with React's hook naming conventions?

Tasneemmohammed0 (approved):
Looks good to me. 🙂  I just left a minor comment.

galasa-ui/src/components/test-runs/TableDesignContent.tsx
James-Cocker: Done 👍

galasa-ui/src/components/common/UseNotification.tsx
James-Cocker: Done 👍

PR #184, opened 2025-08-05
Support Long Test Name in Results Table and Test Run Details Pages

### LINKED ISSUE #2354: Web UI: Clarify Test Name Display and Support Optional Full Name in Results Table
## Describe the bug
The Test name field in the search criteria requires the full test name (the fully-qualified Java class name, e.g., `dev.galasa.zos.ivts.cemt.CEMTManagerIVT`). However, in the Test name column of the results table, we only display the short test name (e.g., `CEMTManagerIVT`). This creates confusion for users who expect that copying the displayed name into the search criteria should work.

## Steps to reproduce
1. Go to Results tab
2. Copy the value shown in the Test name column (e.g., `CEMTManagerIVT`).
3. Paste it into the Test name field in the search criteria and save.
4. Observe that no results are returned.

## Expected behavior
- The Test name column in the results table should be labeled more clearly, e.g., "Test name (short)".
- Users should have the option to view the full test name (including the package) in the results table, but it should not be shown by default.
- In the search criteria, it's acceptable that the user must enter the fully-qualified test name, but the UI should clarify that requirement.

## Tasks
- [x] Text in the test name field to explain it should be fully-qualified with any package detail if there is any.
- [x] Change the name of the existing field "Test name" to "Test name (short)". This is shown by default.
- [x] New field in the table "Test name" This field shows the fully-qualified test name. It is hidden by default.

Tasneemmohammed0:
## Why?
Closes [#2354](https://github.com/galasa-dev/projectmanagement/issues/2354)


galasa-ui/messages/de.json
```diff
@@ -236,7 +236,8 @@
   },
   "OverviewTab": {
     "bundle": "Bundle",
-    "test": "Test",
+    "testName": "Testname",
+    "testShortName": "Testname (short)",
```
techcobweb: I think we need a space in the label "Test name (short)"

Tasneemmohammed0: "Test Name" is correctly translated into german. I just missed the "short" part. 😅

galasa-ui/messages/de.json
```diff
@@ -299,6 +300,7 @@
     "bundle": "Bundle",
     "package": "Paket",
     "testName": "Testname",
+    "testShortName": "Testname (kurz)",
```
techcobweb, line 305: Doesn't match the one above `(short)` or `(kurz)` should be consistent.
"Test name" needs a space in-between too no ?

techcobweb (approved):
Some minor comments, otherwise looks good.

galasa-ui/messages/de.json
Tasneemmohammed0: Fixed. I've corrected the German translation for "short". ✅

PR #186, opened 2025-08-08
Improved Timeframe Filtering, Relative-to-Now Support & Testing

### LINKED ISSUE #2341: Web UI: Test Runs query timeframe can be relative to 'now'
## Story
As a Galasa user I want to define a query which shows runs from the current time into the past, relative to 'now' (whatever that is at the time the query is made)... which I can re-use later to see a similar dataset using the latest 'now'.

ie: To support queries I can save like these:
- my test runs from today (last 24 hours)
- my test runs from this wee (last 7 days)
- all test runs in the past 10 minutes
- test runs marked as BVT tests which failed in the past 3 days
- ...etc...

## Background

Something like this for exact time envelopes:

Image

## Tasks
- [x] Use "Relative to current time" as the radio button text.
- [x] Risk that we can't update values if the control is disabled. eg: Changing the duration when someone picks a different date for `from` or `to`
- [x] `now` radio-button added
- [x] `from` and `to` time grayed-out when `now` selected
- [x] `duration` greyed-out when exact times are used
- [x] When `from` and `to` change, `duration` changes even if it's greyed-out.
- [x] When relative-to-now selected, and duration is changed, the 'from' and 'to' fields update even though they are greyed. (as the current time would have changed).
- [x] url contains `now` with duration rather than an exact `from` and `to` times.
  - [x] set the url when things change
  - [x] parse the url when it has a 'now'
- [x] Every minute, the 'from' and 'to' times change when the relative-to-now radio button is selected. The user is sat on this dialog, so we don't want the times to get out of date.
- [x] When using exact times, the 'to' time can be in the future.
- [x] When using exact times, the 'from' time cannot be more recent than the 'to' time. If someone changes the 'from' time to be more recent than the current 'to' time, then we change the 'to' time to be 1 minute after, with a strip-line/notification carbon info control:
"You selected a 'from' time which is more recent than the 'to' time. This is not a valid time period. The 'to' time has been changed to be slightly more recent than your selected 'from' time."
- [x] The label "show the most recent test runs from 2 days, 1 hour and 13 minutes ago" is shown when the exact times are used. 
- [x] This same text can be used in the results tab

Tasneemmohammed0:
## Why?
Closes [#2341](https://github.com/galasa-dev/projectmanagement/issues/2341)

## Changes
**Exact From Time <=> Exact To Time**
- [x] "To" time can be set in the future.
- [x] When either date is adjusted, the duration updates accordingly, even if it's disabled.

**Duration <=> Exact To Time**
- [x] When the duration changes, the "from" time is recalculated.
- [x] When the  "to" date changes, the duration remains fixed and the  "from" date adjusts.

**Exact From Time <=> Time Relative to Now**
- [x] When "Now" is selected, the "to" date jumps to the current time.
- [x] Any change in the "from" date is reflected to the duration.
- [x] Timeframe text changes to "Showing the most recent test runs from 18 days, 3 hours and 18 minutes ago" when "Now" is selected. 
- [x] Show a warning when "from" date exceeds the current time.

**Duration <=> Time Relative to Now**
- [x] When "Now" is selected, the "to" date jumps to the current time.
- [x] Any change in the duration updates the "from" date, even if the duration field is disabled.
---
**Testing**
- [x] Update `TimeFrameFilter` component tests to match the new flow.
- [x] Add unit tests for the new `DurationFilter` component.
- [x] Add unit tests for the  "duration" param in the `internal-api.test` route.
- [x] Add unit tests  for the "duration" param in the `TestRunsTabs` component.
- [x] Update `TimeFrameContent` component tests to reflect the updated logic, including a dedicated unit test for each of the above cases.


galasa-ui/messages/en.json
```diff
@@ -357,10 +359,15 @@
     "envelopeDescription": "Test runs submitted within this envelope will be shown in the results, subject to other filters being applied.",
     "invalidTimeFrame": "Invalid Time Frame",
     "autoCorrection": "Auto-Correction",
-    "toBeforeFrom": "'To' date cannot be before 'From' date.",
+    "toBeforeFromAutoAdjust": "You selected a 'from' time which is more recent than the 'to' time. This is not a valid time period. The 'to' time has been changed to be slightly more recent than your selected 'from' time.",
```
eamansour: Very minor - when we mention "from" and "to", can these be changed to match the case of "From" and "To" (as that's what the sections are titled)?

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -189,8 +206,18 @@ export default function TestRunsTabs({
     params.set(TEST_RUNS_QUERY_PARAMS.COLUMNS_ORDER, columnsOrder.map((col) => col.id).join(','));
 
     // Timeframe
-    params.set(TEST_RUNS_QUERY_PARAMS.FROM, timeframeValues.fromDate.toISOString());
-    params.set(TEST_RUNS_QUERY_PARAMS.TO, timeframeValues.toDate.toISOString());
+    console.log(timeframeValues.relativeToNow);
```
eamansour: Is this log statement needed?

galasa-ui/src/components/test-runs/timeframe/TimeFrameContent.tsx
```diff
@@ -61,18 +78,32 @@ export const calculateSynchronizedState = (
 export function applyTimeFrameRules(
   fromDate: Date,
   toDate: Date,
+  relativeToNow: boolean | undefined,
```
eamansour: Minor - maybe this could be `isRelativeToNow` to make it clear that it's a boolean elsewhere in the code?

galasa-ui/src/components/test-runs/timeframe/TimeFrameContent.tsx
```diff
@@ -61,18 +78,32 @@ export const calculateSynchronizedState = (
...
+   If the 'from' date is after the 'to' date:
+   - If the query is relative to now, show a warning without adjusting the 'to' time,
+       since the user may want to save a query for the future.
+   - Otherwise, set the 'to' time to one minute after the 'from' time.
+  */
   if (correctedFrom > correctedTo) {
     return {
       correctedFrom: fromDate,
-      correctedTo: toDate,
+      correctedTo: relativeToNow ? toDate : new Date(fromDate.getTime() + 60 * 1000), // 1 minute later
```
eamansour: We could use a constant for the `60 * 1000` part to avoid hard-coding values here

galasa-ui/src/components/test-runs/timeframe/TimeFrameContent.tsx
```diff
@@ -121,59 +149,152 @@ export default function TimeFrameContent({ values, setValues }: TimeFrameContent
...
       <div>
         <p>{translations('selectEnvelope')}</p>
         <p>{translations('envelopeDescription')}</p>
       </div>
-      <TimeFrameFilter values={values} handleValueChange={handleValueChange} />
+
+      <FormGroup className={styles.formGroup} legendText="" role="radiogroup">
+        <div className={styles.fromContainer}>
+          <h3 className={styles.heading}>From</h3>
+          <div className={styles.optionRow}>
```
eamansour: When using dark mode and going to the timeframe tab, the "From" and "To" titles are rendered in white:

Screenshot 2025-08-11 at 14 58 10


Also, the titles here should be internationalized instead of hard-coding `From` and `To`

galasa-ui/messages/en.json
```diff
@@ -357,10 +359,15 @@
...
+    "toBeforeFromWarningOnly": "'From' time is set to the future. Current query will return no results.",
     "dateRangeExceeded": "Date range cannot exceed {maxMonths} months; 'To' date has been adjusted.",
     "dateRangeCapped": "Date range was capped at the current time.",
-    "invalidTimeText": "Please enter a valid time in HH:MM format."
+    "invalidTimeText": "Please enter a valid time in HH:MM format.",
+    "specificTimeTitle": "A specific time",
+    "durationTitle": "Duration before 'To' time",
```
eamansour: The story shows the 'To' part of this message to be in bold, though I don't think this will render in bold.

eamansour (approved):
Looks good! Thanks for making those changes!

PR #189, opened 2025-08-12
Make Test Runs Results Table the Default Tab

### LINKED ISSUE #2361: Web UI: Test runs results table should be the default tab to see when the tab isn't specified in the URL
## Story
As a Galasa tester, I want to see the test results table whenever I enter the test runs part of the web site, based on the current default query, or the default-default query if there isn't one, so that the results I'm most likely to want to see are already in front of me when first landing on that page.

## Background
Landing in the `time frame` tab by default isn't ideal. Now that we have a concept of a single default query, lets use that and show the results immediately.

If there is no tab in the URL, jump to the results table immediately.

## Tasks
- [x] If there is no tab in the URL, jump to the results table immediately.

Tasneemmohammed0:
## Why?
Closes [2361](https://github.com/galasa-dev/projectmanagement/issues/2361)

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunsTabs.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunsTabs.tsx
@@ -74,7 +74,7 @@ export default function TestRunsTabs({
   const [selectedIndex, setSelectedIndex] = useState(() => {
     const tabParam = searchParams.get('tab');
     const initialIndex = tabParam ? TABS_IDS.indexOf(tabParam) : -1;
-    return initialIndex !== -1 ? initialIndex : 0;
+    return initialIndex !== -1 ? initialIndex : 3;
   });
 
   // Initialize selectedVisibleColumns  based on URL parameters or default values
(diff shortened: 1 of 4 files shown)


galasa-ui/src/components/test-runs/TestRunsTabs.tsx
```diff
@@ -74,7 +74,7 @@ export default function TestRunsTabs({
   const [selectedIndex, setSelectedIndex] = useState(() => {
     const tabParam = searchParams.get('tab');
     const initialIndex = tabParam ? TABS_IDS.indexOf(tabParam) : -1;
-    return initialIndex !== -1 ? initialIndex : 0;
+    return initialIndex !== -1 ? initialIndex : 3;
```
eamansour: Instead of hard-coding the index, could we perhaps do something like:

```ts
TABS_IDS.indexOf(<results tab here>)
```

eamansour (approved):
Just one comment but approved assuming it'll be addressed

galasa-ui/src/components/test-runs/TestRunsTabs.tsx
Tasneemmohammed0: Good catch! It’s resolved. :+1:

eamansour (approved):

PR #190, opened 2025-08-12
3270 Tab - Table for Screenshot Selection

James-Cocker:
# Why?
https://github.com/galasa-dev/projectmanagement/issues/2322. This PR Adds the "3270 Terminals" table.


eamansour (dismissed):

galasa-ui/messages/de.json
```diff
@@ -260,6 +260,24 @@
     "recentRunsLink": "Zeige letzte Ausführungen desselben Tests...",
     "runRetriesLink": "Zeige alle Versuche dieses Testlaufs..."
   },
+  "3270Tab": {
+    "Terminal": "Terminal",
+    "ScreenNumber": "Bildschirmnummer",
+    "Time": "Zeit",
+    "Method": "Methode",
+    "searchPlaceholder": "Suchen",
+    "ariaLabel": "ariaLabel",
```
eamansour: Is this translation for ariaLabel needed? Or should the value be updated to something else?

Same applies for the English translations

galasa-ui/public/static/example-terminal-screenshot.png
eamansour, line 1: Do we need this image file checked in if the code that uses it is commented out?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/DisplayTerminalScreenshot.tsx
```diff
@@ -0,0 +1,26 @@
...
+}: {
+  imageData: any;
+  isLoading: boolean;
+}) {
+  return (
+    // <img
+    //   className={styles.screenshot}
+    //   src={`/static/example-terminal-screenshot.png`}
+    //   alt="Placeholder image"
+    // />
```
eamansour: Not a big fan of checking commented-out code, is this still needed?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/DisplayTerminalScreenshot.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+'use client';
+import React, { useEffect, useMemo, useState } from 'react';
+import styles from '@/styles/test-runs/test-run-details/tab3270.module.css';
+
+export default function DisplayTerminalScreenshot({
+  imageData,
+  isLoading,
+}: {
+  imageData: any;
```
eamansour, line 15: Using `any` isn't ideal since it bypasses the type system and defeats the purpose of using TypeScript - can imageData be typed?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const [searchTerm, setSearchTerm] = useState('');
+  const [selectedTerminal, setSelectedTerminal] = useState<DropdownOption | null>(null);
+
+  let screenshotsCollected: boolean = false;
+
+  const handleRowClick = (runId: string, screenshotId: string) => {
+    const newImageData: TerminalImage = allImageData.find(
+      (image) => image.id === screenshotId
+    ) as TerminalImage;
+    console.log('Table row clicked! Image data:\n' + JSON.stringify(newImageData));
```
eamansour: I think this log statement can be removed if it's not needed?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+        Object.values(row).some((value) =>
+          String(value).toLowerCase().includes(lowerCaseSearchTerm)
+        )
+      );
+    }
+
+    if (selectedTerminal && selectedTerminal.id !== 'all') {
+      result = result.filter((row) => row.Terminal === selectedTerminal.id);
+    }
+    // console.log("Hello\n" + JSON.stringify(result));
```
eamansour: Remove comment?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+      };
+
+      fetchData();
+    }
+  }, []);
+
+  if (isLoading) {
+    return (
+      <DataTableSkeleton
+        data-testid="loading-table-skeleton"
```
eamansour: Minor - to avoid mixing test code with actual code it would be better to avoid using data-testid as much as possible - can this be a normal id attribute instead?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+      }) => (
+        <TableContainer>
+          <TableToolbarContent>
+            <TableToolbarSearch
+              placeholder={translations('searchPlaceholder')}
+              persistent
+              onChange={(e: any) => {
+                setSearchTerm(e.target.value);
+              }}
+              data-testid="search-input"
```
eamansour: Same here about avoiding data-testid unless you have to use it

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+            <Dropdown
+              className={styles.terminalDropdownMenu}
+              label="Select a terminal"
+              items={terminalnames}
+              itemToString={(item: DropdownOption | null) => (item ? item.label : '')}
+              onChange={({ selectedItem }: { selectedItem: DropdownOption | null }) => {
+                setSelectedTerminal(selectedItem);
+              }}
+              selectedItem={selectedTerminal}
+              data-testid="terminal-input"
```
eamansour: data-testid here as well

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {rows.map((row) => (
+                <TableRow
+                  key={row.id}
+                  {...getRowProps({ row })}
+                  onClick={() => handleRowClick(runId, row.id)}
+                  className={styles.clickableRow}
+                  data-testid="table-row"
```
eamansour: data-testid here too

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
```diff
@@ -0,0 +1,729 @@
...
+  });
+
+  test('test split with correct input with more than 1 dash in odd placement', () => {
+    expect(splitScreenAndTerminal('term1--1')).toEqual({ screenNumber: 1, terminalName: 'term1-' });
+  });
+});
+
+describe('get3270Screenshots -> populateFlattenedZos3270TerminalDataAndAllImageData', () => {
+  // Example given is only a single image.
+  const mockImages: any = [
```
eamansour: I don't think we need to use `any` here?

Same applies to all the other mock inputs in this file

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+          <TableToolbarContent>
+            <TableToolbarSearch
+              placeholder={translations('searchPlaceholder')}
+              persistent
+              onChange={(e: any) => {
+                setSearchTerm(e.target.value);
+              }}
+              data-testid="search-input"
+            />
+            <Dropdown
```
eamansour, line 240: Running this locally, I noticed that the dropdown appears even if there's only one terminal in a test. The options are "All" and "term1" - is that the expected behaviour? I feel like it would make more sense to only show the dropdown if there are multiple terminals to filter by.

Also, the dropdown's content seems to appear behind the actual table:

Screenshot 2025-08-28 at 13 12 57

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
```diff
@@ -0,0 +1,729 @@
...
+      ],
+    },
+  ];
+
+  beforeEach(() => {
+    flattenedZos3270TerminalData.length = 0;
+    allImageData.length = 0;
+  });
+
+  test('should handle images with an empty fields array', () => {
```
eamansour: These tests seem to be testing functions from get3270Screenshots.ts, should they be moved to a separate test file so that this file focuses solely on testing the TableOfScreenshots component?

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
```diff
@@ -0,0 +1,729 @@
...
+        <TableOfScreenshots
+          isLoading={false}
+          setIsLoading={jest.fn()}
+          setImageData={jest.fn()}
+          {...defaultProps}
+        />
+      );
+    });
+
+    screen.debug();
```
eamansour: I think this screen debug can be removed now?

galasa-ui/src/tests/components/test-runs/test-run-details/ArtifactsTab.test.tsx
```diff
@@ -8,10 +8,11 @@ import { render, screen, fireEvent, waitFor, act } from '@testing-library/react'
 import '@testing-library/jest-dom';
 import { TreeNodeData } from '@/utils/functions/artifacts';
 import { ArtifactsTab } from '@/components/test-runs/test-run-details/ArtifactsTab';
-import { checkForZosTerminalFolderStructure } from '@/utils/checkFor3270FolderStructure';
+import { checkForZosTerminalFolderStructure } from '@/utils/3270/checkFor3270FolderStructure';
 import { downloadArtifactFromServer } from '@/actions/runsAction';
 import { handleDownload } from '@/utils/artifacts';
 import { FeatureFlagProvider } from '@/contexts/FeatureFlagContext';
+import { mock } from 'node:test';
```
eamansour: I don't think `mock` import is being used?

galasa-ui/src/utils/3270/get3270Screenshots.ts
```diff
@@ -0,0 +1,117 @@
...
+import { CellFor3270, TerminalImage, TerminalImageField } from '@/utils/interfaces/common';
+import pako from 'pako';
+
+export const flattenedZos3270TerminalData: CellFor3270[] = [];
+export const allImageData: TerminalImage[] = [];
+
+export function splitScreenAndTerminal(input: string) {
+  const parts = input.split('-');
+
+  if (parts.length < 2 || !Number.isInteger(parseInt(parts[parts.length - 1]))) {
```
eamansour: It might be worth adding a comment to explain why we're splitting on `-` and what the expected format of the `input` string is.

It's not immediately obvious what `input` actually refers to, can this be a more descriptive parameter?

galasa-ui/src/utils/3270/get3270Screenshots.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // 2. Decompress the Uint8Array using pako.
+  const decompressedUint8Array = pako.inflate(gzippedUint8Array);
+
+  // 3. Convert the decompressed Uint8Array back to a string.
+  return JSON.parse(new TextDecoder().decode(decompressedUint8Array));
+}
+
+export function populateFlattenedZos3270TerminalDataAndAllImageData(images: any[]): void {
+  images.forEach((image: any) => {
```
eamansour, line 58: `any` returns! Can `images` be typed so that `image` and then `imageField` can also be typed later on in this function?

galasa-ui/src/utils/constants/common.ts
```diff
@@ -155,7 +155,7 @@ const PREFERENCE_KEYS = {
   TIME_ZONE: 'timeZone' as const,
 } as const;
 
-const TEST_RUN_PAGE_TABS = ['overview', 'methods', 'runLog', 'artifacts'];
+const TEST_RUN_PAGE_TABS = ['overview', 'methods', 'runLog', 'artifacts', 'overview'];
```
eamansour, line 158: Should this be "3270" instead of another "overview"?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
```diff
@@ -0,0 +1,62 @@
...
+  let theme: 'light' | 'dark';
+  if (current === 'system') {
+    if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+      theme = 'dark';
+    } else {
+      theme = 'light';
+    }
+  } else {
+    theme = current;
+  }
```
eamansour, line 43: Do we need to calculate the theme here? I thought the theme would be applied via a global property so we wouldn't need to constantly add `lightTheme` and `darkTheme` CSS classes everywhere

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
```diff
@@ -0,0 +1,215 @@
...
+        getHeaderProps: (options: any) => TableHeadProps;
+        getRowProps: (options: any) => TableRowProps;
+        getTableProps: () => TableBodyProps;
+      }) => (
+        <TableContainer>
+          <TableToolbarContent>
+            <TableToolbarSearch
+              placeholder={translations('searchPlaceholder')}
+              persistent
+              onChange={(e: any) => {
```
eamansour: Can `e` be typed?

eamansour (commented):
Looking good! Just a few comments and suggestions :)

galasa-ui/messages/de.json
James-Cocker: Done 👍 Removed the translation.

galasa-ui/public/static/example-terminal-screenshot.png
James-Cocker: I was planning on leaving the image there until I finished the JSON -> image translation so that I have a reference of what it'll look like and what classnames to give it. The reason I commented this out was because there is no need for the placeholder to be there on main, it's simply to help with the next piece of development.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/DisplayTerminalScreenshot.tsx
James-Cocker: Same as previous comment 👍

galasa-ui/src/components/test-runs/test-run-details/3270Tab/DisplayTerminalScreenshot.tsx
James-Cocker: The reason I have image data as an "any" type is because i do not have enough examples of what sort of fields could be in the data. This is because I need to make sure that I can pick out the correct fields such as row and column without type errors for users with different JSON. I will ensure this is strongly typed before the story is finished by adding this as a task on the 3270 webui story this PR is going towards, will that be okay for now?

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Indeed it can, happy to remove it but i left it there in case anyone wanted to see the onclick functionality. I'll remove it now 👍.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Done, I believe that was in support of the last bit of development for the sticky headers so it must have slipped my last check.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Unfortunately there is no screen.getById() method. I just took another look through the screen.getBy... methods and it doesn't look like anything would be able to get the DataTableSkeleton. I was hoping to use get by role however it doesn't look like the carbon ui item has a role.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Done 👍. Managed to replace the getByTestId() with a getByRole() and remove this data-testid.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Done 👍. One of the elements inside the dropdown has a role of "combobox". Not too sure why it has that name but it seems to be unique within the screen.debug() so i've managed to remove this data-testid as well.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: The <tr> does not seem to have any other unique attributes. I could simply identify that the text appears, however i need to get the specific row element as this test is testing the onClick() method for that specific HTML element.

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
James-Cocker: Same as the previous comment, once I've got a comprehensive set of examples that can be used in the image JSON, I'll be sure to take these off being typed with any in my next PR.

galasa-ui/public/static/example-terminal-screenshot.png
eamansour: Sounds good!

galasa-ui/src/components/test-runs/test-run-details/3270Tab/DisplayTerminalScreenshot.tsx
eamansour: Sure, happy to keep it as-is for now given that the story isn't finished just yet

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
eamansour: Could we perhaps convert it into a `role` attribute and then find by role instead (not sure if that's an attribute on this component)? That would be more useful than a testid - happy to keep this as-is if that's not possible!

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: It was initially the expected behaviour, however i do agree that it makes little sense to keep it way. I've now implemented it in a nice one line change.

I've pinpointed the reason of the dropdown content being underneath the table content to be the stickyHeader prop on the DataTable, hence why it slipped past. I've spent a good 2 hours now attempting different combinations of z-index and position attributes, scanning through the DOM to pinpoint other generated classnames and still cannot find a solution. Do you think we might be able to have a quick chat about it?

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
James-Cocker: I was thinking the same, thank you for bringing this back to my attention. Done. I've also moved the checkFor3270FolderStructure.ts checks out of ArtifactsTab.test.tsx to its own file in the utils folder in tests.

galasa-ui/src/tests/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.test.tsx
James-Cocker: Done 👍

galasa-ui/src/tests/components/test-runs/test-run-details/ArtifactsTab.test.tsx
James-Cocker: Done 👍

galasa-ui/src/utils/3270/get3270Screenshots.ts
James-Cocker: Good points, fixed.

galasa-ui/src/utils/3270/get3270Screenshots.ts
James-Cocker: Same as before, the first step of my next PR will be research into the JSON translation and working out a typed interface for image data with all its optional fields. This will require more examples of the data which i do not believe can be setup currently.

galasa-ui/src/utils/constants/common.ts
James-Cocker: I ran into the problem that the 3270 tab is only shown based on the output of checkFor3270FolderStructure, which caused complications with simply setting the tab - of which i cannot entirely remember now. This is a temporary fix that simply redirects the user to the overview page, and will be tackled properly in the PR for the task `"Perma-link to a particular panel should be copyable to get back to a particular panel. As they all have a unique index number.
eg: https://...../runs/798787868978789?tab=3270&terminal=2&screen=45"`

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
James-Cocker: I found this odd as well. The reason i implemented this logic was because the scroll bar does not seem to comply with the global css theme. An example of this can be seen in the "run log" tab, where even in dark mode, the scroll bar appears in light theme. Would you like me to implement this logic in the run log tab as well?

eamansour: Sounds like this is a wider CSS issue that needs investigating in that case - it's worth raising a defect for it.

As for this logic, I don't think this is the right solution to get around this issue since it means that we'd have to duplicate the logic whenever a scrollbar doesn't comply with the selected theme.

I would either keep this logic temporarily or remove this theme logic for now and tackle the scrollbar issue separately to also fix the issue in the run log tab.

The `lightTheme` and `darkTheme` classes in the `tab370.modules.css` file can also be removed in favour of the global theme.

galasa-ui/src/utils/constants/common.ts
eamansour: Okay, happy to keep this as-is for now in that case!

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Gotcha, i didn't know that role could be assigned in the tsx - thanks!

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
James-Cocker: Done 👍

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TableOfScreenshots.tsx
James-Cocker: Done 👍

eamansour (approved):

PR #191, opened 2025-08-12
feat: persist and use default items per page setting for test runs results table

### LINKED ISSUE #2364: Web UI: User able to store an "Items per page" preference.
## Story
As a Galasa user I want to set a preference for the "Items per page" on the test runs query table, to save me always having to change the default away from 10 to 40 which fits my screen nicely.

## Background
The test runs page has a results table, which is paginated

The page has an 'items per page' which gets frustrating to change all the time

We need a new section on the settings page:

Image

This preference can be stored in local storage

## Tasks
- [x] Implementation + unit tests + translation + local storage use
- [x] Using the value

Tasneemmohammed0:
## Why?
Closes [https://github.com/galasa-dev/projectmanagement/issues/2364](https://github.com/galasa-dev/projectmanagement/issues/2364)
## Changes
- Added a new “Results table default items per page” section in the My Settings page.
-  Introduced a custom hook to save and retrieve the default page size from `localStorage`.
-  Updated the results table to use the saved preference by default.
- **Tests**
  - Unit tests for the new hook.
  - Unit tests for the new settings component.
  - Tests ensuring the results table retreives its default page size from the hook.

### DIFF
--- a/galasa-ui/src/app/mysettings/page.tsx
+++ b/galasa-ui/src/app/mysettings/page.tsx
@@ -19,6 +19,7 @@ import { HOME } from '@/utils/constants/breadcrumb';
 import { fetchUserFromApiServer } from '@/actions/userServerActions';
 import ProfileRole from '@/components/users/UserRole';
 import DateTimeSettings from '@/components/mysettings/DateTimeSettings';
+import ResultsTablePageSizingSetting from '@/components/mysettings/ResultsTablePageSizingSetting';
 export default async function MySettings() {
   const apiConfig = createAuthenticatedApiConfiguration();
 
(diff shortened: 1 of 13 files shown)


galasa-ui/src/app/mysettings/page.tsx
```diff
@@ -65,6 +66,7 @@ export default async function MySettings() {
       />
       <TokenResponseModal refreshToken={refreshToken} clientId={clientId} onLoad={deleteCookies} />
       <DateTimeSettings />
+      <ResultsTablePageSizeSetting />
```
eamansour, line 69: If the user hasn't checked the box in the "Experimental Features" section to view test runs, I think this section should also be hidden.

Otherwise, it might confuse people to see a setting about test runs but they aren't able to see test runs 😅

Tasneemmohammed0: Change resolved along with its unit test. Thank you! 🫡

eamansour (approved):

PR #192, opened 2025-08-15
Improve time display and recent runs logic

Tasneemmohammed0:
## Why?
Addresses [Slack discussion](https://openmainframeproject.slack.com/archives/C05TCCQDE65/p1755072405090689) regarding improvements to test run time display and link generation logic.
## Changes
- Removed the “max months” rule
- Removed automatic adjustment of the “To” date when it is earlier than the “From” date.
- Updated display to show explicit “From” and “To” dates instead of relative times.
- Updated recent runs link logic to fetch runs of the same test from the last two months.


techcobweb:
Given that duration and now are the defaults, could we please move these up above the exact from and exact to time options, so they are at the top please ? @James-Cocker suggested that yesterday, and it's obvious now that the defaults are duration and now. Odd with them at the bottom now.

Tasneemmohammed0:
> Given that duration and now are the defaults, could we please move these up above the exact from and exact to time options, so they are at the top please ? @James-Cocker suggested that yesterday, and it's obvious now that the defaults are duration and now. Odd with them at the bottom now.

Updated :+1:

James-Cocker (approved):

PR #195, opened 2025-08-21
feat(saved-queries): implement full query lifecycle (save, edit, reorder, default)

Tasneemmohammed0:
##  What's Done
- Hard-coded fallback default query if no saved queries exist (e.g. *“Tests ran in the last 24 hours”*).  
- Collapsible sidebar on the Test Runs page.  
- Save button stores the current query URL in local storage (temporary, will later be stored server-side).  
- Ability to name a query and sync it with the URL.  
- Decode a named query from the URL.  
- Display list of saved queries in the sidebar.  
- Re-order saved queries with drag-and-drop (grab handle).  
- Mark one query as default (star icon replaces grab handle).  
- Display current saved query name above the Test Runs table/tabs.  
- Inline renaming of queries (edit field on the right side of the screen).  
- Search/filter queries in the sidebar list as you type.  
- Add new saved query via “+” button - duplicated names get suffixed with (1), (2), etc.  
- Ensure current query is always reflected in the URL (even if not saved).  
- Show drag handles only on hover.  

> [!NOTE]  
> Introduced a new context with the `TestRunsQueryParams` custom hook that centralizes all **URL ↔ state** logic.  
> This makes query state accessible across the application and will be very useful for future page reload optimization plans.  

---
**Unit Test Coverage**
- [x] **`CollapsibleSideBar`**
  - [x] Drag-and-drop
  - [x] Search and filter
  - [x] Add query to list
  - [x] Render sorted queries
- [x] **`SavedQueriesContext`**
  - [x] Test each function individually
- [x] **`TestRunsDetails`**
  - [x] Name editing
  - [x] Saving a query
  - [x] Title and page render
- [x] **`QueryItem`**
- [x] **`TestRunsQueryParamsContext`**
---

## To Be Fixed in a Separate PR
- Sidebar height decreases when the page is empty.    (Edited) Fixed in [#197](https://github.com/galasa-dev/webui/pull/197)
- Scrollbar should appear when the query list exceeds screen height.  

---

## To Be Implemented (Separate PRs)
- Menu button in each query row to allow:  
  - Rename  
  - Delete  
  - Set as default  
- **Important:**
 When clicking on a saved query, the **page should not reload**.  
 Instead, update the **state and URL internally** so filters are applied without a full reload.
  (Edited) Fixed in [#197](https://github.com/galasa-dev/webui/pull/197)
---

##  To Be Finalized
- UI design still evolving - further improvements expected based on feedback.  
>[!Note]
> `createdAt` is currently being used as the saved query ID. This is only a temporary solution until we migrate to a server-side implementation, where each query will have a unique ID stored in the database.


techcobweb:
This is a big PR. It would have been better to deliver it in pieces.
Like :
- The slide-out drawer
- The list of queries
- Drag-drop to re-order them
- Renaming a query
- Search queries...etc.
All would have been easier to review if delivered separately. Worth thinking about for your next story/project I guess.

Tasneemmohammed0:
> This is a big PR. It would have been better to deliver it in pieces. Like :
> 
>     
Yes, I figured that out while working on it, so there wasn’t really any going backward this time. But I’ll definitely take care of it in the upcoming stories.

eamansour:
Hi @Tasneemmohammed0, the changes are looking really good! However, having tried the changes out locally, I'm seeing some strange behaviours that I'll highlight below and I'll also mention some other thoughts I had:

1. Adding a new query then clicking on that query in the "Saved Queries" sidebar will show the old name after the page reloads. When you then try to edit the name of the query, both an error and a success message appears, and another query gets created with the original name! See video below:

https://github.com/user-attachments/assets/22781ad0-4484-48df-96d2-efacfa081c76

2. After creating a query and then editing it (e.g. changing the timeframe), a placeholder success message appears like the one below:
 
Screenshot 2025-08-26 at 10 15 30

3. The "Save Query" button is always active even if no changes have been made to the query. I think it would be better if the button was only made active if changes have been made to the query

4. This can be addressed separately since it's more of a design issue: I don't think it's immediately obvious that you can create a new query. You need to open the sidebar (which I think is missing an indication that the saved queries are in there) and then add the current query in there. Because of this, it's really easy to accidentally change the default query.

5. When you edit the name of a query and then hit 'Enter' or click out of the input field, no success message appears to say that the query was updated/saved - see video below:

https://github.com/user-attachments/assets/114de86a-7289-4700-bd70-d56ea3754782

Let me know if any of these don't make sense!

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -51,8 +67,104 @@ export default function TestRunsDetails({
...
+        kind: 'success',
+        title: translations('successTitle'),
+        subtitle: translations('queryUpdatedMessage', { name: nameToSave }),
+      });
+      setTimeout(() => setNotification(null), NOTIFICATION_VISIBLE_MILLISECS);
+      return;
+    }
+
+    // If the query doesn't exist, create a new one with a unique name
+    const baseName = nameToSave.replace(/\s*\(\d+\)$/, '').trim();
```
eamansour: Regex can be a pain to understand and maintain - can this `baseName` be calculated without regex?

If not, then maybe isolate the regex in a const and explain what the pattern does?

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -77,13 +189,60 @@ export default function TestRunsDetails({
...
+                <input
+                  ref={inputRef}
+                  type="text"
+                  value={editedName}
+                  onChange={(e) => setEditedName(e.target.value)}
+                  onBlur={handleFinishEditing}
+                  onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => {
+                    if (e.key === 'Enter') {
+                      handleFinishEditing();
+                    }
```
eamansour, line 214: The success notification doesn't appear when you finish editing

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -77,13 +189,60 @@ export default function TestRunsDetails({
...
+                  }}
+                  className={styles.queryNameInput}
+                  data-testid="query-name-input"
+                />
+              ) : (
+                <h3 className={styles.queryNameHeading} data-testid="query-name">
```
eamansour: I'm not a big fan of using `data-testid` in actual component code, since it mixes information used only for testing with the production code. Perhaps these tags should be given normal IDs and then you can use query those IDs in test code?

Same applies throughout this PR

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -77,13 +189,60 @@ export default function TestRunsDetails({
...
+                onClick={handleStartEditingName.bind(null, queryName)}
+                size="md"
+              />
+            </div>
+
+            <Button
+              kind="primary"
+              type="button"
+              onClick={handleSaveQuery}
+              data-testid="save-query-button"
```
eamansour: Same here, I really don't like `data-testid` in component code 😅

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -0,0 +1,233 @@
...
+    const currentQuery = savedQueries.find((q) => q.createdAt === active.id);
+    if (currentQuery) {
+      setActiveQuery(currentQuery);
+    }
+  };
+
+  const handleDragEnd = (event: DragEndEvent) => {
+    // Clear the active query state to remove the overlay
+    setActiveQuery(null);
+    const { active, over } = event;
```
eamansour, line 84: `active` and `over` aren't very descriptive, can these be clarified a bit?

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const sensors = useSensors(
+    useSensor(PointerSensor),
+    useSensor(TouchSensor),
+    useSensor(KeyboardSensor, {
+      coordinateGetter: sortableKeyboardCoordinates,
+    })
+  );
+
+  const handleDragStart = (event: DragStartEvent) => {
+    const { active } = event;
```
eamansour, line 70: `active` isn't very descriptive so it's not clear what it actually contains - can this be clarified a bit?

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const handleAddCurrentQuery = () => {
+    const nameToSave = queryName.trim();
+    if (!nameToSave) return;
+
+    const currentUrlParams = new URLSearchParams(searchParams);
+    let finalQueryTitle = nameToSave;
+
+    // Ensure unique query name
+    if (isQuerySaved(finalQueryTitle)) {
+      const baseName = nameToSave.replace(/\s*\(\d+\)$/, '').trim();
```
eamansour: Same comment here regarding using regex which makes code difficult to maintain

galasa-ui/src/components/test-runs/saved-queries/QueryItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <div
+      ref={setNodeRef}
+      style={style}
+      className={`${styles.sideNavItem} ${disabled ? styles.disabled : ''} ${isCollapsed ? styles.collapsed : ''}`}
+    >
+      {isDefault ? (
+        <StarFilled size={18} className={styles.starIcon} />
+      ) : (
+        <Draggable size={18} className={styles.dragHandle} {...attributes} {...listeners} />
```
eamansour: Minor - the `size` value can be stored in a const and referred to twice here so that any changes to the size will be reflected in both icons

galasa-ui/src/hooks/useTestRunsQueryParams.tsx
eamansour, line 1: Empty file - can this be removed?

galasa-ui/src/styles/test-runs/saved-queries/QueryItem.module.css
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  gap: 0.5rem;
+  align-items: flex-start;
+  padding: 0.5rem 0;
+  position: relative;
+  transition: background-color 0.15s ease-in-out;
+}
+
+.sideNavItem.collapsed {
+  display: none !important;
```
eamansour: `!important` isn't great to use since it forcefully overrides values for CSS properties (and in turn messes with specificity) and can make debugging CSS really challenging, so it should be the last thing that you would want to use.

Could `collapsed` be a CSS state? Then it could be used like `.sideNavItem:state(collapsed)`?

galasa-ui/src/styles/test-runs/saved-queries/QueryItem.module.css
```diff
@@ -0,0 +1,72 @@
...
+  word-break: break-word;
+  line-height: 1.4;
+}
+
+.sideNavLink:hover,
+.sideNavLink:focus,
+.sideNavLink:active {
+  text-decoration: none;
+  color: var(--cds-text-primary) !important;
+}
```
eamansour, line 68: Same here regarding `!important` - can these colours be set in a normal way without overriding anything?

galasa-ui/src/styles/test-runs/timeframe/TimeFrameContent.module.css
```diff
@@ -7,7 +7,7 @@
   display: flex;
   flex-direction: column;
   gap: 1.5rem;
-  width: 100%;
+  /* width: 90%; */
```
eamansour: If this is no longer needed, can we remove this instead of commenting it out?

galasa-ui/src/utils/constants/common.ts
```diff
@@ -156,6 +157,13 @@ const PREFERENCE_KEYS = {
 
 const TEST_RUN_PAGE_TABS = ['overview', 'methods', 'runLog', 'artifacts'];
 
+const DEFAULT_QUERY = {
+  // The URL here is the filter state, not the browser URL
+  url: '',
+  title: 'Tests ran in the last 24 hours',
```
eamansour, line 163: Should this default query title be internationalized?

eamansour (commented):
Looking good overall! Just a few comments and suggestions - let me know if anything doesn't make sense!

Tasneemmohammed0:
> Hi @Tasneemmohammed0, the changes are looking really good! However, having tried the changes out locally, I'm seeing some strange behaviours that I'll highlight below and I'll also mention some other thoughts I had:
> 
Hi @eamansour, thanks for the detailed review!
- The first and third points are already addressed in [PR #197](https://github.com/galasa-dev/webui/pull/197) please refer to it.
- I’ve fixed points 2 and 5 in this PR.
- For point 4, I suggest we open a Slack discussion once we’ve had a chance to try the new changes in action.

galasa-ui/src/styles/test-runs/saved-queries/QueryItem.module.css
Tasneemmohammed0: Resolved in [this commit](https://github.com/galasa-dev/webui/pull/197/commits/93e964052c1f90f7e6004f1283355dad95758faf), which is a part of  [197 PR](https://github.com/galasa-dev/webui/pull/197). Since the "Link" Carbon component was removed, there is no longer a need to override any CSS.

galasa-ui/src/utils/constants/common.ts
Tasneemmohammed0: As discussed on Slack, this will be implemented in a separate story.

eamansour (approved):

PR #196, opened 2025-08-24
feat : Always show language selector, remove experimental flag

### LINKED ISSUE #2387: Web UI: Language selection feature to be made un-experimental
## Story
As a Galasa user, I want to see translated languages for the web UI always, rather than  only when I opt-in to an experimental feature.

## Background
This feature is mature enough that we should always show it.

## Tasks
- [ ] The code which checks the experimental feature flag to show/hide the button needs removal.
- [ ] The experimental feature flag for this removed from the settings panel 
Image

- [ ] The icon is a different size from the theme icon next to it. The padding is slightly odd around it which forces the icon to be smaller ?

- [ ] Fix the bug where the menu appears in the wrong place:

https://github.com/user-attachments/assets/3e6077ec-92b9-414d-9d0b-ff1a65654800

shivambisen:
### Description:
- This PR makes the language selector in the Galasa web UI always visible, as the feature is now mature enough to be enabled by default.

---

### Changes include:
- Removed the experimental feature flag check that previously controlled the visibility of the language selector.
- Removed the related flag toggle from the settings panel.
---

closes [#2387](https://github.com/galasa-dev/projectmanagement/issues/2387)


galasa-ui/src/components/headers/LanguageSelector.tsx
```diff
@@ -66,7 +66,7 @@ export default function LanguageSelector() {
           focusTrap={true}
           align="bottom"
           flipped
-          renderIcon={() => <Wikis style={{ fill: 'white' }} />}
+          renderIcon={() => <Wikis style={{ fill: 'white', width: '18px', height: '18px' }} />}
```
eamansour: Can we avoid adding inline styles and keep styling code inside CSS modules instead? That way, we don't mix styling concerns with the rendering logic.

Also another minor issue - can the space before the `:` in the language selector's tooltip be removed? It currently displays as:
```
Selected language : English
```
instead of:
```
Selected language: English
```

eamansour (commented):
One comment but overall looks good! Happy to approve once the comment is addressed.

galasa-ui/src/components/headers/LanguageSelector.tsx
shivambisen: Done moved styles to CSS modules and fixed the extra space in the tooltip

eamansour (approved):

PR #197, opened 2025-08-26
perf(saved-queries): avoid page reloads by managing state internally

Tasneemmohammed0:
## Why? 
As part of [#2340](https://github.com/galasa-dev/projectmanagement/issues/2340) and a follow-up to [#195](https://github.com/galasa-dev/webui/pull/195), this PR introduces navigation between saved queries by managing state internally, eliminating the need for a full page reload.

## Changes
- Added a `lock` to prevent race conditions between URL ↔ state synchronization.
- Updated saved query `Link` to use `replace` for updating the URL on click without triggering a reload (thanks to Next.js).
- Adjusted sidebar height for better layout consistency.


galasa-ui/src/components/test-runs/QueryName.tsx
```diff
@@ -0,0 +1,77 @@
...
+            {isEditingName ? (
+              <input
+                ref={inputRef}
+                type="text"
+                value={editedName}
+                onChange={(e) => setEditedName(e.target.value)}
+                onBlur={handleFinishEditing}
+                onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => {
+                  if (e.key === 'Enter') {
+                    handleFinishEditing();
```
eamansour, line 53: When hitting 'Enter' or clicking away from the input field, the success notification to indicate that the query was saved/updated still doesn't appear. See video to demonstrate this:


https://github.com/user-attachments/assets/39ea636b-29d6-4c3b-8045-42f029b35554

galasa-ui/src/components/test-runs/QueryName.tsx
```diff
@@ -0,0 +1,77 @@
...
+                  }
+                }}
+                className={styles.queryNameInput}
+                data-testid="query-name-input"
+              />
+            ) : (
+              <h3 className={styles.queryNameHeading} data-testid="query-name">
```
eamansour: Can this `data-testid` be removed to avoid mixing test code with production code?

galasa-ui/src/components/test-runs/QueryName.tsx
eamansour: Also, as mentioned in https://github.com/galasa-dev/webui/pull/195#issuecomment-3223384069, the "Save Query" button is always active even if no changes have been made to the query. I think this should be disabled unless there is a change that has been made to the query.

galasa-ui/src/components/test-runs/QueryName.tsx
Tasneemmohammed0: Thanks for spotting that! I had already changed it in the last PR, but merge conflicts have their own magic to mess things up again :) Anyway, I've resolved it.

eamansour (approved):

PR #200, opened 2025-08-29
feat(saved-queries): add full menu actions (edit, delete, default, copy)

Tasneemmohammed0:
## Why? 
As part of [#2340](https://github.com/galasa-dev/projectmanagement/issues/2340) and a follow-up to [#197](https://github.com/galasa-dev/webui/pull/197), this PR introduces full query actions (edit, delete, set as default, copy, duplicate) to improve the saved queries experience.
## Changes
Added a menu button to each query item with the following options:
  - **Rename** => Navigates to the query and focuses on the edit bar.
  - **Set as default** => Sets the selected query as default, unsets the previous default, and moves it to the end of the list.
  - **Duplicate** => Creates a copy of the query’s URL under a unique name (e.g., New Query (1)).
  - **Copy to Clipboard** => Copies the query URL to the clipboard.
  - **Delete** => Removes the query from the list. 
  >[!Note]
   **Delete** and **Set as default** buttons are disabled for the default query since we don't want these operations on the default query.


James-Cocker:
Seems like the "Save Query" button flashes blue before turning grey, any chance this could default to grey on page load? 😁

James-Cocker:
I've also noticed the button changes from grey to blue when no changes are made when flicking between the timeframe and other tabs, is that expected?

Edit: I think I've found why, i think it's because the URL is changing and the save query button changes based on if the URL changes. I believe the saved query should always default to the results page, so is there any way to decode the URL and exclude the current tab section from the encoded URL? This sounds like it might be quite a pain to keep encoding and decoding, so maybe its worth bringing the tab name out as a different parameter, such as "http://localhost:3000/test-runs?q=N4IgLiBcIE4gNCAjgOyiAEgUwDY4PYAEAFAIwCUCIAbgMbowrwxLxhYDOYAygBb4wwAOQCGAWyzwuzDlICCVWvnQc5zJiykBLACbwA5vABG8AA5tOPfoNES2TaWH2yYHKjoCuMEWC3400ADM8AAMoVRw0ABSIjpYAHQAwiIwWij4AAJaRmLxSmIgAL5AA?current-tab=Results". That way, you would not have to change your method to detect changes in the value of the 'q' parameter.

James-Cocker:
One last thing, might it be possible to give the side pannel a max-height so that it has a separate scroll bar?
Screenshot 2025-09-18 at 09 51 39

Tasneemmohammed0:
> I've also noticed the button changes from grey to blue when no changes are made when flicking between the timeframe and other tabs, is that expected?
> 
Yes! I was considering treating tab changes as a new query to be saved. But it makes more sense to have the saved query always open on the results tab, it’s more convenient. Thank you! I’ve fixed it.

Tasneemmohammed0:
> Seems like the "Save Query" button flashes blue before turning grey, any chance this could default to grey on page load? 😁

Fixed :+1:

techcobweb:
One snag: 
image
If you open the menu and it's at the bottom of the page.
I think it should scroll up the page in that case so it's visible ? Or the menu item needs placing somewhere else ?

Lets deal with this separately after this PR is merged.

techcobweb:
Snag: Save a query with "Duration 1,0,0" "now" and jump to a different query with "From explicit time" and "to explicit time" and toggle between the two queries by clicking them on the shortcuts list.

The radio buttons "Duration before To time" and "A specific time" seem to be muddled.

Lets deal with that on a separate PR from this one.

techcobweb:
Snag: Toggle between "Duration before To time" and "A specific time" and the "save Query" button doesn't enable. This is probably related to above.
Lets deal with this on a separate PR.

techcobweb:
Snag: Not sure about this one, but if I go to teh timeframe tab, then copy the URL. Should that URL take me to the Results tab ? I think so, to prepare the way for when the timeframe is inside a dialog behind a button.

techcobweb:
Snag:
Go to the page concerning a single test, hit the `View recent runs of the same test...` link and the title in the results has nothing to do with the name of the testcase ... was hoping it would say "Recent runs of test TestSleep" or similar. 

Lets deal with this in a separate PR.

galasa-ui/src/contexts/TestRunsQueryParamsContext.tsx
```diff
@@ -27,6 +27,7 @@ import {
   MINUTE_MS,
   HOUR_MS,
   TABS_IDS,
+  DEFAULT_QUERY,
```
techcobweb, line 30: Why are we saving the fact that it's the default query ?

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -97,12 +103,15 @@ export default function TestRunsDetails({
...
-    const queryToRename = getQuery(oldName);
+    const queryToRename = getQueryByName(oldName);
 
     // If it was a saved query, perform the rename in storage
     if (queryToRename) {
       const updatedUrlParams = new URLSearchParams(searchParams);
+
+      // Update the URL parameters with the new query name and switch to results tab
       updatedUrlParams.set(TEST_RUNS_QUERY_PARAMS.QUERY_NAME, newName);
+      updatedUrlParams.set(TEST_RUNS_QUERY_PARAMS.TAB, TABS_IDS[3]);
```
eamansour, line 114: This won't be an issue when we move away from using tabs in the test runs page, so I'm fine with keeping this as-is, but I would generally avoid hard-coding specific index values like `TABS_IDS[3]` here in case the values in `TABS_IDS` change.

Same applies on this line https://github.com/galasa-dev/webui/pull/200/files#diff-49b4051c3eb5283a7d47e18390a2d7519673ea3ddd41c03fe1ffbf8a50e5f635R143 and https://github.com/galasa-dev/webui/pull/200/files#diff-3b273c22cdbca9ac15c6915d2fe3bee582dd79d4487f71a1e6a5b80893f6f9ccR114

galasa-ui/src/utils/constants/common.ts
```diff
@@ -4,7 +4,9 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 
+import { encode } from 'punycode';
 import { ColumnDefinition } from '../interfaces';
+import { encodeStateToUrlParam } from '../urlEncoder';
```
eamansour: Unused imports here?

eamansour:
A few more snags, but I think they can all be done separately:

1. Can be fixed in a separate PR: The "Add current query" tooltip appears behind the main page content, so only "Add current" shows up:

Screenshot 2025-09-30 at 15 52 42

2. Like 2 can also be fixed in a separate PR, the "Actions" tooltip is slightly cut off on the right:
Screenshot 2025-09-30 at 16 01 25

3. Can be addressed separately: the default "Tests ran in the last 24 hours" query seems to use the specific "from" time, so it won't always show the tests ran in the last 24 hours as time passes. Should it default to using "now" and duration from 1 day to always be accurate?

4. More minor visual issues: having the web UI on a small browser window causes the query name to wrap around and become difficult to use and the page content seems to flow past the header/footer width (see second image below):

Screenshot 2025-09-30 at 16 15 57

Screenshot 2025-09-30 at 16 19 47

5. Duplicating the default query doesn't seem to save the new query. The "Save Query" button is enabled after duplicating. However, for non-default queries, the "Save Query" button is disabled after duplicating, which is good.

6. Hovering over the "Delete" button for the default query makes the "Delete" text change colour, I think it should stay the same colour as the button is disabled. The "Delete" text also becomes white in light mode for the default query.
Screenshot 2025-09-30 at 16 35 10

Tasneemmohammed0:
> Snag: Go to the page concerning a single test, hit the `View recent runs of the same test...` link and the title in the results has nothing to do with the name of the testcase ... was hoping it would say "Recent runs of test TestSleep" or similar.
> 
> Lets deal with this in a separate PR.

We’re already addressing this in #[207](https://github.com/galasa-dev/webui/pull/207)

galasa-ui/src/contexts/TestRunsQueryParamsContext.tsx
Tasneemmohammed0: Since we’re using the default query in multiple places for comparisons, it makes sense to define it as a constant rather than duplicating it.

eamansour (approved):

PR #205, opened 2025-09-22
James C 3270 url sharing

James-Cocker:
# Why?

Adds the URL sharing functionality for 3270 tabs in the form ```http://localhost:3000/test-runs/cdb-20a9f606-f70e-403e-9429-8b5f246894b1-1758348015432-C26431?tab=3270&terminalScreen=IYK2ZNB5_1-1``` where `IYK2ZNB5_1-1` is terminal "IYK2ZNB5_1" screen "1"


galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
```diff
@@ -31,25 +35,28 @@ export default function TabFor3270({
...
-  }
+  }, [highlightedRowId]);
+
+  // Get the 'terminalScreen' parameter
+  useEffect(() => {
+    if (is3270CurrentlySelected && highlightedRowId === '') {
+      const url = new URL(window.location.href);
+      setHighlightedRowId(url.searchParams.get('terminalScreen') || '');
+    }
+  }, [is3270CurrentlySelected]);
```
eamansour, line 50: I believe these useEffects will throw up warnings about missing dependencies. The first one will complain about `is3270CurrentlySelected` and the second will complain about `highlightedRowId`, can these warnings be addressed?

galasa-ui/src/components/test-runs/search-criteria/SearchCriteriaContent.tsx
```diff
@@ -317,7 +318,7 @@ export default function SearchCriteriaContent({
     filterableFields.find((field) => field.id === selectedFilterId) || filterableFields[0];
 
   return (
-    <div>
+    <div className={`${GetTheme() === 'light' ? styles.lightTheme : styles.darkTheme}`}>
```
eamansour: I'm still not quite sold on applying a lightTheme/darkTheme class here - if we're applying a global theme based on the current theme selection, why do we still need to assign a theme-specific CSS class to this div? Are we not able to use the global theme?

If we want to override the theme for this specific component, we could use a Carbon <Theme> component like we do in the breadcrumb component

galasa-ui/src/components/test-runs/test-run-details/LogTab.tsx
```diff
@@ -566,7 +567,9 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
     : translations('selectLinesToCreatePermalink');
 
   return (
-    <div className={styles.tabContent}>
+    <div
+      className={`${styles.tabContent} ${GetTheme() === 'light' ? styles.lightTheme : styles.darkTheme}`}
```
eamansour: Same comment here regarding applying lightTheme/darkTheme CSS classes instead of using the global CSS applied based on the currently-selected theme.

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
```diff
@@ -31,25 +35,28 @@ export default function TabFor3270({
...
 
   if (isError) {
     return <ErrorPage />;
   }
 
   return (
     <div
-      className={`${styles.tab3270Container} ${theme === 'light' ? styles.lightTheme : styles.darkTheme}`}
+      className={`${styles.tab3270Container} ${GetTheme() === 'light' ? styles.lightTheme : styles.darkTheme}`}
```
eamansour: Same comment here regarding applying lightTheme/darkTheme CSS classes instead of using the global CSS applied based on the currently-selected theme.

galasa-ui/src/components/test-runs/test-run-details/TestRunDetails.tsx
```diff
@@ -74,18 +74,33 @@ const TestRunDetails = ({
...
+      const tabName = searchParams.get(SINGLE_RUN_QUERY_PARAMS.TAB)!;
+
+      // Redirect 3270 tab to overview page until it has been verified that the test has a 3270 folder structure populated with images
+      if (tabName === '3270') {
+        setIs3270TabSelectedInURL(true);
+        return TEST_RUN_PAGE_TABS.indexOf('overview');
+      }
+      return TEST_RUN_PAGE_TABS.indexOf(tabName);
+    }
+    return 0;
```
eamansour: Really minor suggestion - this code could be easier to read if we had a single return statement at the end of this useState function. For example, we could do a:

```ts
let tabIndex = 0;
```

then we could update the value wherever we need it to change, like:

```ts
if (tabName === '3270') {
  ...
  tabIndex = TEST_RUN_PAGE_TABS.indexOf('overview');
} else {
  tabIndex = TEST_RUN_PAGE_TABS.indexOf(tabName);
}
return tabIndex;
```

galasa-ui/src/styles/global.scss
```diff
@@ -89,47 +89,3 @@ $spaceSizes: (1, 2, 3, 5, 8, 13);
...
-}
-
-::-webkit-scrollbar-thumb {
-  background-color: var(--scrollbar-thumb);
-  border-radius: 4px;
-}
-
-::-webkit-scrollbar-thumb:hover {
-  background-color: var(--scrollbar-thumb-hover);
-}
```
eamansour, line 135: This was added as a global fix to stop the scrollbars in the web UI on different browsers from appearing with a white background. Is there a reason why this was removed?

galasa-ui/src/styles/test-runs/TestRunsPage.module.css
```diff
@@ -351,3 +351,11 @@
...
   cursor: pointer;
 }
+
+.lightTheme {
+  color-scheme: light;
+}
+
+.darkTheme {
+  color-scheme: dark;
+}
```
eamansour: See comments about using the global theme rather than creating `lightTheme` and `darkTheme` CSS classes for individual components. I think this isn't needed.

galasa-ui/src/styles/test-runs/test-run-details/LogTab.module.css
```diff
@@ -107,3 +106,11 @@
...
   cursor: not-allowed;
 }
+
+.lightTheme {
+  color-scheme: light;
+}
+
+.darkTheme {
+  color-scheme: dark;
+}
```
eamansour: See comments about using the global theme rather than creating `lightTheme` and `darkTheme` CSS classes for individual components. I think this isn't needed.

galasa-ui/src/components/GetTheme.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import React from 'react';
```
eamansour: Unused import?

eamansour (approved):
A few comments, mostly minor but happy to approve assuming they'll be addressed.

galasa-ui/src/components/test-runs/search-criteria/SearchCriteriaContent.tsx
James-Cocker: Indeed it is an issue, but only for the colour theme of the scrollbars. Initially, Mike looked into globally applying a separate theming - however this introduced some very strange padding issues with my 3270 screenshots table (as seen in yesterday's scrum i believe). The reason the global theming is not applied to the scrollbars by carbon is due to a known issue which is currently being addressed by the CarbonUI team which Mike found - so hopefully we can delete this code soon. It's not ideal, but it looks like we'll have to do this for each scrollable component for now.

galasa-ui/src/components/test-runs/test-run-details/TestRunDetails.tsx
James-Cocker: Great point! I'll do that now.

galasa-ui/src/styles/global.scss
James-Cocker: Yes - whilst a global theme would have been much preferred, it seems the only way to get these scroll bars to have the correct css applied to them is with `color-scheme` attribute set - and i'm not sure if this can be done on a global scale.

galasa-ui/src/components/test-runs/search-criteria/SearchCriteriaContent.tsx
eamansour: Thanks for the explanation - I sent that CarbonUI issue to Mike when he was adding the global scrollbar styling, not sure when it'll be fixed unfortunately. However, if the padding issues were caused by the properties in the globally-applied CSS, could that have been fixed by modifying those properties? For example, if there are any that interact with width/height, then maybe they could be removed since we're only interested in changing the colour of the scrollbars, rather than their dimensions?

Interestingly, I don't see the issue on main, and I'm wondering if the issue was because Mike's browser always displays scrollbars (if that's possible). Do you see the padding issue on prod1?

I'm just concerned that the more we add to the web UI, the more technical debt we're adding with this code being everywhere where scrollbars are involved.

If there's no other way to solve this, then it would be worth raising a story to make sure we don't lose track of this technical debt.

galasa-ui/src/styles/global.scss
eamansour: Would this work?

```css
/* Default scrollbar styling (light theme) */
:root {
  color-scheme: light;
}

/* Dark theme scrollbar styling */
[data-carbon-theme='dark'] {
  color-scheme: dark;
}
```

galasa-ui/src/components/test-runs/search-criteria/SearchCriteriaContent.tsx
eamansour: See https://github.com/galasa-dev/webui/pull/205#discussion_r2390788830 - that seems to work for me locally

galasa-ui/src/components/GetTheme.ts
James-Cocker: There was an issue where the GetTheme function wasn't being identified as a component as it wasn't in the components folder - but i added this import as part of the troubleshooting - removing now

galasa-ui/src/styles/global.scss
James-Cocker: Works perfectly - I seem to have a knack for overcomplicating some of these things. I'll be sure to keep this in mind for future, thanks!

galasa-ui/src/styles/test-runs/test-run-details/LogTab.module.css
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/components/GetTheme.ts
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/styles/test-runs/TestRunsPage.module.css
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/styles/global.scss
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/components/test-runs/test-run-details/LogTab.tsx
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/components/test-runs/search-criteria/SearchCriteriaContent.tsx
James-Cocker: Removed GetTheme method and all mentions of it in favour of a 4 line change in global.scss

galasa-ui/src/components/test-runs/test-run-details/3270Tab/TabFor3270.tsx
James-Cocker: You're completely right about the first one, I've added is3270CurrentlySelected to the dependency array - however if i added highlightedRowId to the dep array for getting the terminalScreen parameter, it means it would be getting the new row id from the URL then immediately setting it again to the URL which is redundant. Not sure if there's a way around this warning.

PR #207, opened 2025-09-25
Fix Test Runs Page title not updating when viewing recent runs

### LINKED ISSUE #2402: Web UI: Test Runs Page title does not update when viewing recent runs of a test
## Describe the bug
When viewing recent test runs of a specific test run, the page title still shows the previous query name. The title should instead describe the content currently being displayed.

## Steps to reproduce
1. Go to any Single Test Run page.
2. Click on the "View recent runs of the same test..." link
3. Observe that the title still shows the previous query name.

## Expected behavior
The page title should update to reflect the context, e.g.:“All attempts of test run C12345”.

## Additional context

https://github.com/user-attachments/assets/7ab0461a-6478-4611-b811-b651a11eb9d8

Tasneemmohammed0:
## Why
Closes [#2402](https://github.com/galasa-dev/projectmanagement/issues/2402)
## Changes
 - Added a `queryName` parameter to the single test run page links to explain what the current query is.
 - Updated unit tests.


galasa-ui/src/utils/functions/translations.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import { TranslationValues } from 'next-intl';
+import { TranslationFn } from '../types/common';
+
+/**
+ * Function to get the translation for a given key, and if not found, return the key itself.
```
techcobweb: Would it be better to return the English variant if the German is missing ? And some log failure in the console log or elsewhere ?
And the key only if the English is missing ?

galasa-ui/src/utils/urlStateMappers.ts
```diff
@@ -25,6 +25,7 @@ const keyMap: Record<string, string> = {
   [TEST_RUNS_QUERY_PARAMS.COLUMNS_ORDER]: 'co',
   [TEST_RUNS_QUERY_PARAMS.SORT_ORDER]: 'so',
   [TEST_RUNS_QUERY_PARAMS.QUERY_NAME]: 'qn',
+  [TEST_RUNS_QUERY_PARAMS.PREVIOUS_TEST_RUN_NAME]: 'ptrn',
```
techcobweb: I'm happy if the name of the test run is just set when it's launched. From that point on, we shouldn't know how to distinguish it against a normal query. ie: If the user is in German when they hit the link, they get a german test run appear. This is fine. Seems that you are storing the type of query in the URL now so it can be displayed in a different way ? That will cause a trouble when someone wants to edit that query no ?

Tasneemmohammed0: Update - As discussed on Slack, I rolled back the translations and we’ll sort them out later.

eamansour (approved):
Looks good! Thanks for making these changes!

PR #217, opened 2025-10-16
James C 3270 Add Cursor Colour Inversion 

James-Cocker:
# Why?

As the cursor position is not displayed, inverting the colours of the cursor's position seems to be the best way to display its whereabouts.

image

Note: the cursor might actually be in a different place for this specific image as the logic for the backend has not been pushed into main yet

### DIFF
--- a/galasa-ui/src/utils/3270/screenshotRenderer.ts
+++ b/galasa-ui/src/utils/3270/screenshotRenderer.ts
@@ -68,7 +68,6 @@ function drawCharacters(
         if (characterCell.cursor) {
           // Invert background and foreground
           context.fillStyle = characterColour;
-          console.log(characterWidth);
           context.fillRect(x, y - cellHeight / 2, characterWidth, cellHeight);
           context.fillStyle = backgroundColour;
           context.fillText(characterCell.character, x, y);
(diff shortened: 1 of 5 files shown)


galasa-ui/src/utils/3270/screenshotRenderer.ts
```diff
@@ -65,7 +65,17 @@ function drawCharacters(
     // Edit context if character cell found.
     row.forEach((characterCell) => {
       if (characterCell) {
-        context.fillText(characterCell.character, x, y);
+        if (characterCell.cursor) {
+          // Invert background and foreground
+          context.fillStyle = characterColour;
+          console.log(characterWidth);
```
eamansour: If it's not needed, we can remove this console.log message

eamansour (approved):
Code looks good, just one tiny comment around removing a log message if it's not needed.

For future reference, it would be great if you could add a screenshot to the PR description to show visual changes!

galasa-ui/src/utils/3270/screenshotRenderer.ts
James-Cocker: Good spot - removed 👍

PR #218, opened 2025-10-17
James C 3270 screenshots still being fetched multiple times

James-Cocker:
# Why?

Fixes bug which caused multiple GET requests for the same test run when navigating back and forward to a table of screenshots. Now, the previous set of results is cached correctly. Note: reloading the page will still force the GET requests to be re-done, but if the user uses the browser navigation to go backwards and forwards on a page of screenshots, or navigates back and clicks on the same test run then the cached results are used.


galasa-ui/src/utils/3270/get3270Screenshots.ts
```diff
@@ -107,32 +108,39 @@ export function populateFlattenedZos3270TerminalDataAndAllImageData(images: Term
...
+    newFlattenedZos3270TerminalData.length === 0 ||
+    previousRunId !== runId
+  ) {
+    newAllImageData.length = 0;
+    newFlattenedZos3270TerminalData.length = 0;
+
+    for (var terminal of zos3270TerminalData) {
+      const zippedFilesContainingImageJSON: FileNode[] = Object.values(terminal.children)
+        .filter((node) => (node as FileNode).isFile)
+        .map((node) => node as FileNode);
```
eamansour: Minor - could the `map` and `filter` operations be swapped around so that we don't need the `as FileNode` bit twice?

Something like this:

```suggestion
        .map((node) => node as FileNode)
        .filter((node) => node.isFile);
```

eamansour (approved):

galasa-ui/src/utils/3270/get3270Screenshots.ts
James-Cocker: Great idea - done now thank you

PR #226, opened 2025-10-28
Dev Container

James-Cocker:
# Why?

Adds a dev container to the Galasa webui for faster setup for new contributors.


.devcontainer/dev-container-setup.sh
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright contributors to the Galasa project
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+function check_required_environemnt_variables () {
```
eamansour: Typo - environemnt

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  "remoteUser": "node",
+
+  "mounts": [
+    "source=${localEnv:HOME}/.m2,target=/home/node/.m2,type=bind,consistency=cached",
+    "source=${localEnv:NODE_EXTRA_CA_CERTS},target=/home/node/certs.pem,type=bind,consistency=cached"
+  ],
+
+  // Note: changing your local environment variables requires a vscode reset.
```
eamansour: Minor - replace "reset" with "restart" ? We don't want people to reset their vscode settings

README.md
```diff
@@ -3,33 +3,73 @@
...
+### GALASA_DEV_TOKEN and GALASA_API_SERVER_URL
+- In order to [connect the local development web UI with a remote API server](#connecting-the-local-development-web-ui-with-a-remote-api-server), you must set the following environment variables (see link for instructions).
+
+- `GALASA_DEV_TOKEN`: a token that the web UI uses in development mode to connect to a remote service.
+
+- `GALASA_API_SERVER_URL`: the URL of your running Galasa service, e.g. `https://my-galasa-service.com/api`.
+
+### NODE_EXTRA_CA_CERTS and NODE_USE_SYSTEM_CA
+
+- You are required to validate yourself when connecting to a Galasa service. This can be done with `NODE_EXTRA_CA_CERTS` or `NODE_USE_SYSTEM_CA` (only available to the manual approach).
```
eamansour: NODE_EXTRA_CA_CERTS and NODE_USE_SYSTEM_CA aren't required if the galasa service you're connecting to doesn't use certificates signed by unknown certificate authorities (CAs) like corporate certificates.

"you are required to validate yourself" doesn't quite make sense - it could be improved to say that if the remote Galasa service that you want to connect to uses certificates signed by internal or corporate CAs, then you'll need to set either of these environment variables so that the webui can trust the certificates.

README.md
```diff
@@ -3,33 +3,73 @@
...
 
-## To build locally
+- `NODE_EXTRA_CA_CERTS`: The path to a PEM file which should contain your intermediate and root CA certificates, e.g. `/Users/user_name/Galasa/certs.pem`. 
 
-To build the code, it needs to get the REST interface definition from somewhere, to generate stubs in order to make calling the REST API easy.
-Hence, we need to tell the build where to get this definition from. 
+- `NODE_USE_SYSTEM_CA` (only available to the manual approach): This tells the UI to ignore certificate errors and trust server certificates if they are in the local system keychain, e.g. `NODE_USE_SYSTEM_CA=1`. This is OK for development purposes, but it is not advised to use this technique on a production deployment of a UI.
```
eamansour: The NODE_USE_SYSTEM_CA env variable doesn't ignore certificate errors, so it's worth removing that part

.devcontainer/dev-container-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+    REQUIRED_ENVS=("GALASA_API_SERVER_URL")
+
+    for VAR in "${REQUIRED_ENVS[@]}"; do
+        if [ -z "${!VAR}" ]; then
+            echo "Error: Required environment variable $VAR is not set! Please set the variable on your host machine and restart VS Code."
```
eamansour: Someone can set the webui's environment variables via a .env.development.local file, which gets loaded when the webui starts up, so I don't think this script is actually needed.

Someone can also just set the environment variable in the terminal within the devcontainer, so you don't have to set it in your host machine.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+  "containerEnv": {
+    "GALASA_DEV_TOKEN":"${localEnv:GALASA_DEV_TOKEN}",
+    "GALASA_API_SERVER_URL":"${localEnv:GALASA_API_SERVER_URL}",
```
eamansour: Is there a way for the GALASA_DEV_TOKEN and GALASA_API_SERVER_URL env variables to only be set if a value has been set in the host machine?

When I tried the devcontainer locally, I didn't have either set since I use a .env.development.local file within the webui that gets loaded. The devcontainer had set these to have empty values (e.g. `GALASA_DEV_TOKEN=""`) which prevented the values from my .env.development.local to be picked up.

To fix this, I had to manually run `unset GALASA_DEV_TOKEN` and `unset GALASA_API_SERVER_URL` and then it picked up the values from the .env.development.local file.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+    "SOURCE_MAVEN":"${localEnv:SOURCE_MAVEN}",
+    "NODE_EXTRA_CA_CERTS":"/home/node/certs.pem"
+  },
+
+  "forwardPorts": [3000],
+
+  "postCreateCommand": "bash -c './.devcontainer/dev-container-setup.sh && ./build-locally.sh -c'"
```
eamansour: Not sure if this is the same for you, but when I open the devcontainer and run a build-locally.sh, the `npm install` and jest tests are really slow!

Is there any way you could speed the build up in the devcontainer?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+  "mounts": [
+    "source=${localEnv:HOME}/.m2,target=/home/node/.m2,type=bind,consistency=cached",
+    "source=${localEnv:NODE_EXTRA_CA_CERTS},target=/home/node/certs.pem,type=bind,consistency=cached"
+  ],
+
+  // Note: changing your local environment variables requires a vscode reset.
+  "containerEnv": {
+    "GALASA_DEV_TOKEN":"${localEnv:GALASA_DEV_TOKEN}",
+    "GALASA_API_SERVER_URL":"${localEnv:GALASA_API_SERVER_URL}",
+    "SOURCE_MAVEN":"${localEnv:SOURCE_MAVEN}",
```
eamansour: Similar issue for the SOURCE_MAVEN env variable - if the user hasn't set SOURCE_MAVEN and hasn't built galasa locally, then the build-locally script will not default this the maven repo at development.galasa.dev because it'll see `SOURCE_MAVEN=""`

I think we should only set this env variable into the devcontainer if the host machine has the environment variable set to a non-empty value

.devcontainer/dev-container-setup.sh
James-Cocker: Of course! Great point, the error logs might not display the best errors but with the updated readme, and the possibility of a local env file it does seem this script is rather pointless - especially for only checking one variable. 

Note: I set it up as a for loop as i mistakingly thought that other variables such as the CA certificates were required if you were curious why i did implemented it as a loop.

README.md
James-Cocker: Gotcha! I went with this: `If you want to connect to a remote Galasa service that requires certificates signed by internal or corporate CAs, then you'll need to set either NODE_EXTRA_CA_CERTS or NODE_USE_SYSTEM_CA (manual set up only) so that the frontend can trust the certificates.`

README.md
James-Cocker: I see, i grabbed that from the 'common problems' section which i believe Mike wrote so I'll update that as well 👍

.devcontainer/devcontainer.json
James-Cocker: I see, I believe I attempted to do that with the SOURCE_MAVEN variable and it seemed to not have a system of doing it in the devcontainer.json but i might be able to set the variables within a script and pass them in as parameters

James-Cocker: Update: It simply required a small update to the syntax as such: `"GALASA_DEV_TOKEN":"${localEnv:GALASA_DEV_TOKEN?}",` (extra question mark)

.devcontainer/devcontainer.json
James-Cocker: Seems like mounting the node_modules folder as a named volume might be of help, will attempt after lunch

eamansour (approved):

PR #227, opened 2025-10-28
Correct Skeleton Image Styling on 3270 Tab

James-Cocker:
# Why?

Ensures skeleton loading image is desired width and height.

### DIFF
--- a/galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
+++ b/galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
@@ -61,11 +61,11 @@
   max-height: min(calc(100vh - 450px), 550px);
 }
 
-.skeletonScreenshot {
-  object-fit: contain !important;
-  max-width: 1100px !important;
-  height: calc(100% - 48px) !important;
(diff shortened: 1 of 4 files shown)


galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
```diff
@@ -62,10 +62,10 @@
...
 
 .skeletonScreenshot {
-  object-fit: contain;
-  max-width: 1100px;
-  height: calc(100% - 48px);
-  width: auto;
+  object-fit: contain !important;
+  max-width: 1100px !important;
+  height: calc(100% - 48px) !important;
+  width: auto !important;
```
eamansour: I’m really not a big fan of using `!important` in CSS since it overrides the specificity of properties - it could come back to bite us in the future if we continue to use it in different places.

Is there a different way that you could fix the skeleton behaviour? For example, if `!important` worked, then that might mean there's something else that’s setting these properties, so using an even more specific CSS selector (like an ID) for the component could work?

eamansour (approved):

PR #233, opened 2025-11-27
James c saved queries styling changes

James-Cocker:
# Why?

Solves [#2445](https://github.com/galasa-dev/projectmanagement/issues/2445) with various stying changes (see below). Please note it does **not** seem to be possible to remove the tooltips and that the positions are very restrictive, especially with all custom CSS being overridden.

- Moving the `Add current query` tooltip so that the entire message is  shown image
- Shifting the nav bar contents to the left so that the entire tooltip for the overflow menus are visible. image
- Delete button no longer has a text colour change on hover image

### DIFF
--- a/galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
+++ b/galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
@@ -172,18 +172,17 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
             <div className={styles.innerContentWrapper}>
               <p className={styles.headerTitle}>{translations('savedQueries')}</p>
               <div className={styles.toolbar}>
-                <div id={styles.collapsibleSideNavSearch}>
-                  <Search
-                    labelText={translations('searchSavedQueries')}
-                    placeholder={translations('searchSavedQueries')}
-                    size="lg"
-                    value={searchTerm}
-                    onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
-                      setSearchTerm(e.target.value)
-                    }
-                    onClear={() => setSearchTerm('')}
-                  />
-                </div>
+                <Search
+                  className={styles.collapsibleSideNavSearch}
+                  labelText={translations('searchSavedQueries')}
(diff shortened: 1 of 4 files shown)


jadecarino:
@James-Cocker Hi James, thanks for fixing the "Delete" text changing colour when disabled.

It certainly looks better that "Add current query" is now being cut off, but I'm not loving the fact that the text isn't centered with the arrow.

In regards to the tooltips, I'm wondering if it would a) be more consistent and b) be a better fix for our problem with "Add current query" cutting off slightly, to just configure all tooltips to come out to the left/right? I'm not sure if this is an unusual thing to do, and I don't know how many tooltips we have in the UI currently if this would be a lot of changes, but just a thought.

@eamansour what are your thoughts?

eamansour:
Just had a thought about fixing this @James-Cocker - one of the sidebar containers likely isn't allowing content inside it to overflow, so setting `overflow: visible` in the CSS might be what we're missing. You might need to work your way through the sidebar's wrappers to see which one needs changing.

If that works, then we could move the tooltips to all appear to the right of the sidebar for consistency like Jade mentions.

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -172,23 +172,26 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
...
+                    labelText={translations('searchSavedQueries')}
+                    placeholder={translations('searchSavedQueries')}
+                    size="lg"
+                    value={searchTerm}
+                    onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+                      setSearchTerm(e.target.value)
+                    }
+                    onClear={() => setSearchTerm('')}
+                  />
+                </div>
```
eamansour: Instead of adding another div, could we not just add a `className={styles.collapsibleSideNavSearch}` property into the `<Search>` component itself and turn `collapsibleSideNavSearch` into a CSS class?

Tried it myself and it seems fine:

Screenshot 2025-12-04 at 12 40 39

eamansour (approved):

PR #234, opened 2025-12-03
Iss2486 - Allow users to open a test run into a new tab

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2486

## Changes

- Allow users to right click a row in the Test Runs table and open the test run in a new tab:
   - Embed an anchor element into each cell in the Test Runs table. The anchor element has an href attribute of the URL for the test run page associated with that cell's row. This will allow users to right click on any cell in a row, and choose to copy the link or open that link in a new tab if they wish. 
   - Two new style classes added to the CSS file for the cells and the link overlay from the anchor element.
   - Unit test to validate that every table cell has a link with the correct href to its run.
- Push the 'Test Runs' breadcrumb to the trail on load of a 'Test Run: xxx' page, instead of on click of a row on the Test Runs table. This is so the breadcrumb is added to the trail both when the row is clicked, and if a user opens a new tab, to go to a Test Run page.
   - Unit tests updated.

### DIFF
--- a/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
+++ b/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
@@ -73,7 +73,6 @@ export default function TestRunsTable({
   durationMinutes,
 }: TestRunsTableProps) {
   const translations = useTranslations('TestRunsTable');
-  const { pushBreadCrumb } = useHistoryBreadCrumbs();
   const { formatDate } = useDateTimeFormat();
 
   const router = useRouter();
@@ -131,12 +130,6 @@ export default function TestRunsTable({
 
   // Navigate to the test run details page using the runId
   const handleRowClick = (runId: string, runName: string) => {
-    // Push the current page URL to the breadcrumb history
-    pushBreadCrumb({
-      ...TEST_RUNS,
-      route: `/test-runs?${searchParams.toString()}`,
-    });
-
     // Navigate to the test run details page
     router.push(`/test-runs/${runId}`);
   };
--- a/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
+++ b/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
@@ -40,6 +40,7 @@ import { useDateTimeFormat } from '@/contexts/DateTimeFormatContext';
 import { useDisappearingNotification } from '@/hooks/useDisappearingNotification';
 import { getTimeframeText } from '@/utils/functions/timeFrameText';
 import useResultsTablePageSize from '@/hooks/useResultsTablePageSize';
+import Link from 'next/link';
 
 interface CustomCellProps {
   header: string;
@@ -142,7 +143,7 @@ export default function TestRunsTable({
     let cellComponent = (
       <TableCell className={styles.linkCell}>
         {value}
-        <a href={href} className={styles.linkOverlay} />
+        <Link href={href} prefetch={false} className={styles.linkOverlay} />
       </TableCell>
     );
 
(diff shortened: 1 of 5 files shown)


James-Cocker:
Looks Great! Just to mention one thing, it looks like clicking on it vs opening in a new tab causes a difference in the bread crumb (see below). The expected behaviour is for a test details page for the bread crumb to be `Home / Test Runs`

Test opened in new tab:
image

Test clicked on:
image

Looks like you can give that a fix by adding this code:
```
pushBreadCrumb({
      ...TEST_RUNS,
      route: `/test-runs?${searchParams.toString()}`,
    });
 ```

**Comment by jadecarino** [comment]
@James-Cocker Thanks James, I have hopefully fixed this in https://github.com/galasa-dev/webui/pull/234/commits/f0316b258e9b601094bbae5e995459f379fafabc

File: galasa-ui/src/components/test-runs/results/TestRunsTable.tsx [hunk 1]
```diff
@@ -144,22 +138,33 @@ export default function TestRunsTable({
...
         </TableCell>
       );
     } else if (header === 'submittedAt') {
       // Format the date using the context's formatDate function
-      cellComponent = <TableCell>{formatDate(new Date(value))}</TableCell>;
+      cellComponent = (
+        <TableCell className={styles.linkCell}>
+          {formatDate(new Date(value))}
+          <a href={href} className={styles.linkOverlay} />
+        </TableCell>
```
eamansour, line 167: The `<a>` tags seem to be making the screen flash when clicking on a row, which is more visible when using dark mode.

You could try using Next.js' `<Link>` tag with `prefetch={false}` to avoid pre-fetching all of the test run detail pages. See https://nextjs.org/docs/app/api-reference/components/link

galasa-ui/src/components/test-runs/test-run-details/TestRunDetails.tsx
```diff
@@ -179,6 +180,16 @@ const TestRunDetails = ({
...
 
+  useEffect(() => {
+    // If the 'Test Runs' breadcrumb is already in the items, skip.
+    if (breadCrumbItems.length > 1) return;
+    // Push the Test Runs URL to the breadcrumb history.
+    pushBreadCrumb({
+      ...TEST_RUNS,
+      route: `/test-runs?${searchParams.toString()}`,
+    });
+  });
```
eamansour, line 191: Looks like there's a bit of strange behaviour happening with the breadcrumbs now:

1. Click on a test run from the table
2. Use your browser's back button
3. Notice "Test runs" is still in the breadcrumbs even though you're back to the test runs page already

galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
jadecarino: Thanks Eamonn, great suggestion, I've fixed this in https://github.com/galasa-dev/webui/pull/234/commits/6505638d65b139d36e2e3c65d6d5aa4d47adeaa8

galasa-ui/src/components/test-runs/test-run-details/TestRunDetails.tsx
eamansour: Discussed on slack - Jade will raise a separate defect to fix the wider issue with the breadcrumbs not appearing correctly when using browser buttons.

jadecarino: Thanks Eamonn, as discussed on Slack, it looks like we have a more general bug around the Breadcrumbs not being correct if you use the browser's back/forward buttons instead of links/buttons on the UI, so I will raise a separate bug for this and fix in a different change set so that we can get this new feature delivered now.

eamansour (approved):

PR #236, opened 2025-12-11
James C fix height of saved queries

### LINKED ISSUE #2492: Webui saved queries: height of saved queries wrapper set to constant 800px
## Describe the bug
In Galasa's saved queries the height is set to a constant 800px no matter the height of the contents on the web page. This means if the text within the table flows onto the next line (due to adding more columns, a smaller screen width etc.) the inner wrapper of the side bar will stick to 800px. Setting overflow to be auto to have a scrollable list of saved queries inherently demands a fixed amount (ie. can't be flex or proportion of parent), so the solution must work around this in some manor.

## Steps to reproduce
1. Go to /test-runs
2. Open saved queries
3. Set number of queries per page on results table to 50
4. Notice the inner wrapper stays consistent at 800px when the screen content grows.

## Expected behaviour
A workaround where the inner wrapper of the saved queries is somehow set to the height of the web page contents (with the size of header, footer etc. subtracted)

## Additional context
Image

Image

James-Cocker:
# Why?
Fixes [#2492](https://github.com/galasa-dev/projectmanagement/issues/2492)

Dynamically sizes the nav bar even though the bar has a scrollable list of items by shrinking it to 700px, taking a snapshot of the height of the main content, and sizes accordingly. Necessary to shrink as the wrapper must be set to a fixed height, like how it was before this PR.


galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -57,12 +58,30 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
   // State to hold the data of the item currently being dragged for the DragOverlay
   const [activeQuery, setActiveQuery] = useState<SavedQueryType | null>(null);
 
+  const [sideNavExpandedHeight, setSideNavExpandedHeight] = useState(0);
+  const [mainContentElement, setMainContentElement] = useState<Element | null>(null);
+  const SIDE_NAV_MIN_HEIGHT = 700;
```
eamansour: Minor - can the `sideNavExpandedHeight` and `SIDE_NAV_MIN_HEIGHT` have the unit of measurement in their names? For example `sideNavExpandedHeightPixels` or `SIDE_NAV_MIN_HEIGHT_PIXELS`?

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -148,6 +167,37 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
...
+      // Check if there's a valid entry.
+      if (entries[0]) {
+        updateSideNavHeight();
+      }
+    });
+
+    if (mainContentElement) {
+      resizeObserver.observe(mainContentElement);
+    } else {
+      setSideNavExpandedHeight(800);
```
eamansour: Minor - can the 800 be in a const as well?

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -148,6 +167,37 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
...
+    }
+
+    // Cleanup function to remove the event listener when the component unmounts
+    return () => {
+      if (mainContentElement) {
+        resizeObserver.unobserve(mainContentElement);
+      }
```
eamansour, line 198: Nice!

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -57,12 +58,30 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
...
+  const [mainContentElement, setMainContentElement] = useState<Element | null>(null);
+  const SIDE_NAV_MIN_HEIGHT = 700;
+
   // Isolate user-sortable queries from the default query
   const sortableQueries = useMemo(
     () => savedQueries.filter((query) => query.createdAt !== defaultQuery.createdAt),
     [savedQueries, defaultQuery]
   );
 
+  const updateSideNavHeight = () => {
```
eamansour: Since this `updateSideNavHeight` is only used in a useEffect, we can define the function inside the useEffect so that we can get rid of the missing dependency warning in the useEffect.

galasa-ui/src/tests/components/test-runs/saved-queries/CollapsibleSideBar.test.tsx
```diff
@@ -40,6 +43,13 @@ jest.mock('@/contexts/TestRunsQueryParamsContext', () => ({
   useTestRunsQueryParams: jest.fn(),
 }));
 
+// Stylesheet seems to somehow override CollapsibleSideBar.module.css, so the extra necessary items have been added in here.
+jest.mock('@/styles/test-runs/TestRunsPage.module.css', () => ({
+  mainContent: 'TestRunsPage_mainContent__Ftan5',
+  sideNavCollapsed: 'sideNavCollapsed',
+  sideNavExpanded: 'sideNavExpanded',
+}));
+
```
eamansour: I don't think this is needed - removing this does cause the `should observe the main content if main content rendered, and set to height of main content -50px` test to fail, but setting the string 'mainContent' into the className seems to make the test pass again:

```ts
    test('should observe the main content if main content rendered, and set to height of main content -50px', async () => {
      const mainContentElement = document.createElement('div');
      mainContentElement.className = 'mainContent';
      document.body.appendChild(mainContentElement);
      ...
```

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -57,12 +58,30 @@ export default function CollapsibleSideBar({ handleEditQueryName }: CollapsibleS
...
+    if (mainContentElement) {
+      // As the mainContent for the test runs details is also flex, we must set this height to a minimum (700), wait a short while, then set the height of this element to the main content minus an offset.
```
eamansour: Minor - can we remove the `(700)` from the comment since that'll be another place we need to remember to update if we ever change the minimum height :)

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
```diff
@@ -5,10 +5,11 @@
  */
```
eamansour, line 5: Just tried the changes out locally and overall it looks really good! I've spotted one snag in it that (likely already existed previously) where the action menu for the very last query in the sidebar can end up appearing below the footer so you're forced to scroll down to see the "Delete" button:

Screenshot 2025-12-11 at 13 23 22

I'll raise a new defect for this so we can tackle this separately.

eamansour (approved):
Looks good! Just a few minor comments but happy to approve :)

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
James-Cocker: Indeed I noticed this too. An easy fix would be to set the direction of the overflow contents to come out the top - thoughts on that?

galasa-ui/src/tests/components/test-runs/saved-queries/CollapsibleSideBar.test.tsx
James-Cocker: Interesting! Great spot I've implemented that now.

galasa-ui/src/components/test-runs/saved-queries/CollapsibleSideBar.tsx
James-Cocker: Thank you!

PR #241, opened 2026-01-06
Iss2389 - Search test runs by user, table can display user, and details view displays user

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2389

## Changes

- Test run details page shows the "User" associated with a run
- Test runs can now be searched by "User" (this works on it's own or when used in combination with "Requestor")
- "User" column can be added to the test runs table
   - This has been set as a default visible column to keep it on par with "Requestor"
- To ensure that pre-existing saved queries created before this feature was delivered can see "User" as a row in the Table Design, a function `syncResultsTableColumnsWithQueryParams` has been added to the urlEncoder which checks if the saved query URL includes `user` in the `columnsOrder` parameter (as this is used to populate the Table Design rows), and if it doesn't, will inject it in
   - This function is general and should behave the same for any future fields we add
- Unit tests that validated the `columnsOrder` parameter have now been updated to include all results table columns, as these tests will pass through the new function `syncResultsTableColumnsWithQueryParams` and so must include all table columns


James-Cocker:
Overall PR looks great to me! Just thought I'd reiterate Eamonn's point of if it's possible to decode the user's set of URLs, check if the user parameter is in there, and if not then adding it. If this is possible, it would be great for it to be generalised for any new field made in the future and add that in before the PR goes ahead. 

I.e. when the saved-queries tab is loaded, decode, look through each of them, check the URL against the defaults and if one of them is **missing**, then inject it in. A similar functionality must be performed when the "reset to default" button checks the current URL against the constructed URL of the default queries.

James-Cocker (approved):

James-Cocker:
Also, one last comment. Might there be a scenario where a user would want to search through only the 'unknown' user kicked off tests? Currently it doesn't seem to be possible to.

jadecarino:
@James-Cocker Thanks for your review and comments. I've now added a function to add the "user" parameter into pre-existing saved queries URLs so that it shows up as a row on the Table Design table. This function checks if any results table columns are missing from the URL and so should work in the same way if we added any more new fields in future.

Re your second comment, I am unsure what you mean by searching for an 'unknown' user. The user will default to the same as the 'requestor' if the tester did not attempt to override the user value, so there should never be a test with an unknown user. Hopefully that answers your question.

eamansour (approved):

James-Cocker:
@jadecarino After some investigating it looks like the the 'user' column was coming up with 'unknown' simply due to the fact i was connecting to my local minikube - so all good from me!

PR #244, opened 2026-01-08
James C iss2466 update tags on existing test run

James-Cocker:
# Why?
Solves iss2466 by adding an edit button on the overview page of a test run. This edit button will allow the user to add or remove tags on an existing test run.


galasa-ui/src/app/api/ras/runs/[runid]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    return NextResponse.json({ success: true, tags: tags });
+  } catch (error: any) {
+    console.error('Error updating run tags:', error);
+    return NextResponse.json(
+      { success: false, error: error.message || 'Failed to update tags' },
+      { status: 500 }
+    );
+  }
+}
```
Comment (eamansour) [inline galasa-ui/src/app/api/ras/runs/[runid]/route.ts]: Given that we're sending a request to mutate data, I think a server action would be more suitable here rather than an API route - see https://nextjs.org/docs/app/getting-started/updating-data

That way, we could just call the server action as a function from the client component and pass in the data we want to send instead of manually building up a request.

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -23,6 +25,16 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
 
   const [weekBefore, setWeekBefore] = useState<string | null>(null);
 
+  const [isTagsEditModalOpen, setIsTagsEditModalOpen] = useState<Boolean>(false);
```
eamansour: Can we use `boolean` instead of `Boolean` here?

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -39,13 +51,100 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+      // Call the API route to update tags using the staged tags Set.
+      const response = await fetch(`/api/ras/runs/${metadata.runId}`, {
+        method: 'PUT',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          tags: Array.from(stagedTags),
+        }),
+      });
```
eamansour: See previous comment about switching to using a server action - if we use a server action, we should be able to import the function and call it here without worrying about building up a HTTP request

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -39,13 +51,100 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+        subtitle: translations('updateSuccessMessage'),
+      });
+
+      // Close modal after a short delay to show success message.
+      setTimeout(() => {
+        handleModalClose();
+
+        // Refresh the page to show the updated tags.
+        window.location.reload();
+      }, 1500);
```
eamansour: Very minor - can this timeout (1500) value be set as a constant?

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -94,6 +196,48 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+              title={notification.title}
+              subtitle={notification.subtitle}
+              lowContrast
+              hideCloseButton={false}
+              onCloseButtonClick={() => setNotification(null)}
+            />
+          )}
+          <TextInput
+            data-modal-primary-focus
+            id="text-input-1"
```
eamansour: Could the ID `text-input-1` be made more specific to this input? Something like `tag-name-text-input`?

galasa-ui/src/components/test-runs/test-run-details/RenderTags.tsx
```diff
@@ -0,0 +1,81 @@
...
+type TagSize = 'sm' | 'md' | 'lg';
+
+const RenderTags = ({
+  tags,
+  dismissible,
+  size,
+  onTagRemove,
+}: {
+  tags: string[];
+  dismissible: boolean;
```
eamansour: Minor - maybe `dismissible` could be `isDismissible`?

galasa-ui/src/utils/functions/textToHexColour.ts
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+const tagColours: [string, string][] = [
```
eamansour, line 11: I still think moving away from using a fixed 2D array would be good in the long-term (especially if we later allow users to set colours themselves) 😅 I'm happy to keep this as-is for now though!

One small improvement we could make here is to use constants for white and black instead of repeating `#FFFFFF` and `#000000`.

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -94,6 +196,48 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+              className={styles.notification}
+              kind={notification.kind}
+              title={notification.title}
+              subtitle={notification.subtitle}
+              lowContrast
+              hideCloseButton={false}
+              onCloseButtonClick={() => setNotification(null)}
+            />
+          )}
+          <TextInput
```
eamansour: I think it would be better to use similar controls like in the "Tags" search criteria rather than a text input, where we can add/remove tags individually. That would make this UI more consistent when managing tags.

After trying it out locally, it feels a bit strange to have a single text input used to add multiple tags. Also, the placeholder text doesn't make it obvious that I have to either use commas or spaces to add multiple tags.

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -94,6 +196,48 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
           )}
         </div>
       </div>
+
+      {isTagsEditModalOpen && (
+        <Modal
```
eamansour: Not sure if this is just me, but the modal's launch animation doesn't seem to play when I click to edit the tags, so the modal just pops into existence.

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -39,13 +51,100 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+        title: translations('updateSuccess'),
+        subtitle: translations('updateSuccessMessage'),
+      });
+
+      // Close modal after a short delay to show success message.
+      setTimeout(() => {
+        handleModalClose();
+
+        // Refresh the page to show the updated tags.
+        window.location.reload();
```
eamansour: Is there a way we could avoid reloading the page to get the updated tags?

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
James-Cocker: Looks like that was a part of the default component I copied over from carbon so I'll go ahead and simply remove this, thank you

galasa-ui/src/utils/functions/textToHexColour.ts
James-Cocker: Very true! I'll do that now

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
James-Cocker: Due to the future improvements around this tag component that we discussed I'll go ahead and close this for now.

eamansour (approved):
Great work!

PR #246, opened 2026-01-14
Set the correct initial state for the timeframe to fix the specific from/to times not working properly, remove unused imports

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2514

Previously, the initial states for the selected "from" and "to" options were being set to the specific times even though the default query uses the "duration" and "now" options - this is because the `isRelativeToNow` value isn't set immediately, so the selected options always initialise to the wrong options.

This PR sets the default selected options to "duration" and "now", while also making sure that the "to" option remains synchronized with the `isRelativeToNow` value so that the correct "to" option is selected when switching between queries.

## Changes
- [x] Set the initial `selectedFromOption` and `selectedToOption` states in the timeframe tab to be `duration` and `now`, respectively
- [x] Removed unused imports from various components for a general tidy-up

### DIFF
--- a/galasa-ui/src/tests/components/test-runs/timeframe/TimeFrameContent.test.tsx
+++ b/galasa-ui/src/tests/components/test-runs/timeframe/TimeFrameContent.test.tsx
@@ -8,8 +8,9 @@ import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import TimeFrameContent, {
   applyTimeFrameRules,
   calculateSynchronizedState,
+  FromSelectionOptions,
+  ToSelectionOptions,
 } from '@/components/test-runs/timeframe/TimeFrameContent';
-import { addMonths } from '@/utils/timeOperations';
 import { DAY_MS } from '@/utils/constants/common';
 import { TimeFrameValues } from '@/utils/interfaces';
 import { useState } from 'react';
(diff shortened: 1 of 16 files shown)


galasa-ui/src/tests/components/test-runs/timeframe/TimeFrameContent.test.tsx
```diff
@@ -8,8 +8,9 @@ import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import TimeFrameContent, {
   applyTimeFrameRules,
   calculateSynchronizedState,
+  FromSelectionOptions,
+  ToSelectionOptions,
 } from '@/components/test-runs/timeframe/TimeFrameContent';
-import { addMonths } from '@/utils/timeOperations';
```
James-Cocker, line 12: Looks like this was the only import/ mention of this function, is it worth keeping or should we de-clutter now?

eamansour: Great spot! I've removed that function in https://github.com/galasa-dev/webui/pull/246/commits/ebbc265995d50e5a221e3d40ac1b9611fc1f0836

James-Cocker (approved):

PR #247, opened 2026-01-19
Tag fetching and searchable tags, but overrides existing list of tags

James-Cocker:
# Why?
Solves [#2466](https://github.com/galasa-dev/projectmanagement/issues/2466) by extending functionality of editing tags to be a filterable, selectable list of tags where you can query existing tag objects.

image

image

(Below shows the multi select component filtering)
image

image

(Below shows tags overflowing on the test-runs page)
image


galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
```diff
@@ -143,7 +143,7 @@ export default function TestRunsTable({
       const tagsArray = value.split(', ');
       return (
         <TableCell className={styles.linkCell}>
-          <RenderTags tags={tagsArray} isDismissible={false} size="sm" />
+          <RenderTags tags={tagsArray.sort()} isDismissible={false} size="sm" />
```
jadecarino: Very minor but could this `sort()` go into the line 143 above where we create the tagsArray? I think we should keep any processing/sorting of the data separate to when it's placed into a component.

galasa-ui/src/components/test-runs/test-run-details/OverviewTab.tsx
```diff
@@ -161,13 +193,25 @@ const OverviewTab = ({ metadata }: { metadata: RunMetadata }) => {
...
+            }}
+            tabIndex={0}
+            role="button"
+            aria-label={translations('editTags')}
           >
             <Edit className={styles.tagsEditButton} />
           </div>
         </h5>
-        <RenderTags tags={tags} isDismissible={false} size="md" />
+        <RenderTags tags={tags.sort()} isDismissible={false} size="md" />
```
jadecarino, line 211: Same comment as previous, could we add the `sort()` to where we do `setTags` if possible?

jadecarino (changes requested):

jadecarino:
@James-Cocker Hi James, from my testing mostly looks good but there's still some funny behaviour with the sorting. Here's how to recreate:
1. Go to a test run with no runs
2. Add the tags 'artifact' and 'http' from the dropdown
3. Add a tag 'jade' (after http alphabetically) and it appears in the correct alphabetical order (artifact, http, jade)
4. Then unselect 'http' as a tag and the order will put 'http' at the bottom of the list

Is this the intended behaviour? I've been playing around and it seems to be some inconsistencies between sometimes the ticked tags group together and then the unticked tags group together, and each group is alphabetically sorted, and then sometimes alternatively the whole tags list both ticked and unticked are combined and sorted alphabetically as a whole.  Whatever behaviour you decide upon should stay consistent I believe.

James-Cocker:
Hey @jadecarino I found the issue that you were experiencing and showed me on our call. Should be all fixed now 😁.

jadecarino (approved):
Looks good, thanks James! Lets hold fire for now on merging until I have decided on when we are doing the patch release.

PR #250, opened 2026-01-22
Iss2485 - Add persistent toolbar to test runs table to allow users to filter table data

jadecarino:
## Why?

Contributes to https://github.com/galasa-dev/projectmanagement/issues/2485

## Changes

- Adds a persistent toolbar to the test runs table to allow users to filter table data with a search text bar.
- The filter will update on each change of the search text, i.e., every new character typed/deleted.
- Users can filter by fields of a run that are currently visible, i.e., if the Bundle is `dev.galasa.zos` and the Test Name is `CICSTest`, but the Bundle is not visible in the table, the filter will match on `CICS` but not `zos`.
- Filter will search across the whole dataset in the table and collapse into the pages sizes, and also update the 'total' number in the pagination bar.
- Table is set back to page 1 when the filter is updated to avoid this situation: a filter with 6 pages of data, the user is on page 6, and then when a new filter is added with only 2 pages of data, the user has to scroll back through 4 empty pages.

Note:
- This is the first part of the story to make it quicker and easier to search by Run Name. Another PR will follow with a new dedicated search box that is visible at the top of the Test Runs page where users can search by Run Name only.

### DIFF
--- a/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
+++ b/galasa-ui/src/components/test-runs/results/TestRunsTable.tsx
@@ -98,32 +98,33 @@ export default function TestRunsTable({
   // Filter rows in the current paginatedRows if the text in the
   // persistent toolbar search box changes and matches any table info.
   const filteredRows = useMemo(() => {
+    if (!runsList || runsList.length === 0) {
+      return [];
+    }
+
     const searchLowerCase = search.toLowerCase();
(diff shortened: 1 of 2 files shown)


James-Cocker:
Summary of thoughts on functionality: Condensing down the data works perfectly, however it looks like the custom search logic slightly breaks down at the datetime and tags fields as it is searching on strings that don't quite match what's visible in the table.

James-Cocker (approved):
Great work!

PR #251, opened 2026-01-27
Iss2485 - Search test runs by run name with static Search box

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2485

## Changes

- [x] New TestRunsSearch component that is available in the TestRunsDetails (table page) and TestRunDetails (details page) that allows users to search for tests by run name.
- [x] If a single run is found, they are routed directly to that test run details page.
- [x] If no run is found, an Info notification tells them.
- [x] If multiple re-runs are found of the same run name, the page will navigate to the latest instance of that run (i.e., the run that managed to finish, regardless of result).
- [x] The Go button attached to the Search component is disabled if the search input is blank.
- [x] New function in the internal-api which calls the RAS API with just a run name lookup instead of calling the pre-existing fetch runs with pagination that is too complex for this feature.
- [x] New jest tests to test the behaviour and rendering of TestRunsSearch.
- [x] Updates to existing jest tests for TestRunsDetails and TestRunDetails to support rendering of a `Search` component (fixes undefined elements).

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunsDetails.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunsDetails.tsx
@@ -7,11 +7,11 @@
 import BreadCrumb from '@/components/common/BreadCrumb';
 import TestRunsTabs from '@/components/test-runs/TestRunsTabs';
 import styles from '@/styles/test-runs/TestRunsPage.module.css';
-import { Suspense, useEffect, useMemo, useRef, useState } from 'react';
+import { FormEvent, Suspense, useEffect, useMemo, useRef, useState } from 'react';
 import useHistoryBreadCrumbs from '@/hooks/useHistoryBreadCrumbs';
 import { useTranslations } from 'next-intl';
 import { NotificationType } from '@/utils/types/common';
-import { Button, InlineNotification } from '@carbon/react';
+import { Button, InlineNotification, Search } from '@carbon/react';
 import { Share } from '@carbon/icons-react';
 import PageTile from '../PageTile';
 import CollapsibleSideBar from './saved-queries/CollapsibleSideBar';
@@ -25,6 +25,8 @@ import {
 import { decodeStateFromUrlParam, encodeStateToUrlParam } from '@/utils/encoding/urlEncoder';
 import QueryName from './QueryName';
 import { generateUniqueQueryName } from '@/utils/functions/savedQueries';
+import { TestRunsData } from '@/utils/testRuns';
+import { useRouter } from 'next/navigation';
 
 interface TestRunsDetailsProps {
   requestorNamesPromise: Promise<string[]>;
@@ -45,9 +47,15 @@ export default function TestRunsDetails({
   const [isEditingName, setIsEditingName] = useState<boolean>(false);
   const [editedName, setEditedName] = useState<string>('');
 
+  const [currentSearchInput, setCurrentSearchInput] = useState('');
+  const [goButtonDisabled, setGoButtonDisabled] = useState(true);
+  const [searchNotification, setSearchNotification] = useState<NotificationType | null>(null);
+
   const inputRef = useRef<HTMLInputElement>(null);
 
   const activeQuery = getQueryByName(queryName);
+
+  const router = useRouter();
 
   // Focus and select the input when editing
   useEffect(() => {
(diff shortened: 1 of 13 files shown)


galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -219,9 +227,91 @@ export default function TestRunsDetails({
...
+  const onSubmit = async (event: FormEvent) => {
+    event.preventDefault();
+
+    const searchRunName = currentSearchInput.toUpperCase();
+    // Search from earliest time to now as timeframe is irrelevant
+    const searchFrom = new Date(Date.UTC(0, 0, 1, 0, 0, 0)).toISOString();
+    const searchTo = new Date().toISOString();
+
+    const url = new URL(
+      `/internal-api/test-runs?runName=${searchRunName}&from=${searchFrom}&to=${searchTo}`,
```
techcobweb: If you are doing an exact match on run name, do you need the time bracket ? That should be an index lookup no matter what the timeframe.

techcobweb (approved):
I don't really like the UI design. Seems that it should be part of the search criteria, and just make that better. 

I think the behaviour is supposed to launch the first run if there are re-run records... but I don't see any unit tests for that.

Should we consider an additional URL which has a simple
testRunName=xxx parameter ? Then the user could just change the URL also ?

Code looks like it should probably do what it was supposed to do though, so approving.

techcobweb:
Seems to me that it should be lower down on the UI:

image

And it should probably be collapsed until you hit the search icon.
like this? 
image

Also, do you need the Go button ? Seems a bit superfluous ? And if so, make it as small as possible ?

techcobweb:
The 'jumping to the first hit without showing table' could be a global part of the search criteria, savable as part of a search criteria...etc.

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
jadecarino: Hi Mike, thank you for your review. I was just using the function that already existed in the internal-api which required from/to times but I have now made a new function that calls the RAS API with just a run name lookup, which my Search component now uses. This is in commit https://github.com/galasa-dev/webui/pull/251/commits/7d275ff8f002513b385352492320328269a29a27

jadecarino:
@techcobweb Response to your comments below:

> I don't really like the UI design. Seems that it should be part of the search criteria, and just make that better.

Yes we do have stories to improve the Search Criteria with the modals which we would like to get to but they will take a little longer. This was just a request from Mark to get in ASAP so they could easily lookup runs by run name with as minimal clicks as possible which this does. I have moved the Search bar lower on the page though as Mark's preference was that it be in the main content panel (the white panel).

> I think the behaviour is supposed to launch the first run if there are re-run records... but I don't see any unit tests for that.

I have provided tests for this in https://github.com/galasa-dev/webui/pull/251/commits/570370cc96481f9cc3ef90b554969da8f6dd3aa4

> Should we consider an additional URL which has a simple testRunName=xxx parameter ? Then the user could just change the URL also?

Definitely something we could consider...

jadecarino:
> Seems to me that it should be lower down on the UI:
> 
> image
> And it should probably be collapsed until you hit the search icon. like this? image [ image ](https://camo.githubusercontent.com/e2f930b265a17a8e5b9d0740396e3aa69546cdd610234ec0a0ec9c7a589eae10/68747470733a2f2f636172626f6e64657369676e73797374656d2e636f6d2f33363763323031326366616633343965333532616664613432383436343565612f646174617461626c655f7365617263682e676966) [ ](https://camo.githubusercontent.com/e2f930b265a17a8e5b9d0740396e3aa69546cdd610234ec0a0ec9c7a589eae10/68747470733a2f2f636172626f6e64657369676e73797374656d2e636f6d2f33363763323031326366616633343965333532616664613432383436343565612f646174617461626c655f7365617263682e676966)
> 
> Also, do you need the Go button ? Seems a bit superfluous ? And if so, make it as small as possible ?

The component you've sent on the screen video is a different Carbon component - the table toolbar. We have added this also (previous PR) to filter the current dataset in the table. The Search component does not collapse down like the table toolbar does.

PR #258, opened 2026-02-11
JamesC iss2481 Run Log Scrolling

### LINKED ISSUE #2481: Web UI: a scroll to bottom + a refresh button would be good when viewing the run log
## Story
As a UI user, I want to often go to the bottom of the run log & then refresh, so that I can see new entries added to in-progress tests.

## Background
Self-explanatory - Additionally, the main window scroll bar doesn't work in harmony with the run log scroll bar, it would be better if the content resized to the full window size, so a scroll bar isn't need for the main window

Image

## Tasks
- [x] Ensure run log container's scroll bar does become hidden
- [x] Add a button to refresh the run log
- [x] Add a button to jump to the top of the run log
- [x] Add a button to jump to the bottom of the run log

James-Cocker:
# Why?

Closes https://github.com/galasa-dev/projectmanagement/issues/2481 with a shortened runLog container, added buttons to refresh and jump to top/ bottom.


galasa-ui/src/actions/runsAction.ts
```diff
@@ -92,3 +93,13 @@ export const getExistingTagObjects = async () => {
...
+
+export const fetchRunLog = async (runId: string) => {
+  let runLog;
+  try {
+    runLog = await fetchRunDetailLogs(runId);
+  } catch (error: any) {
+    runLog = 'Error fetching run log: ' + error;
+  }
+  return runLog;
+};
```
eamansour, line 104: Instead of modifying the run log's content on error, would it be better to keep the old content the same and display a Carbon error notification that tells the user that there was an error fetching the run log? https://carbondesignsystem.com/components/notification/usage/

galasa-ui/src/components/test-runs/test-run-details/LogTab.tsx
```diff
@@ -162,6 +171,58 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
...
+      setIsAtTop(scrollTop <= THRESHOLD_PIXELS);
+      setIsAtBottom(scrollTop + clientHeight >= scrollHeight - THRESHOLD_PIXELS);
+    }
+  }, []);
+
+  const scrollToTop = () => {
+    if (scrollContainerRef.current) {
+      scrollContainerRef.current.scrollTo({
+        top: 0,
+        behavior: 'smooth',
```
eamansour: Regarding accessibility: if someone wants reduced motion, using `smooth` would cause the scroll to slowly move to the top/bottom rather than instantly jumping to the top/bottom - see https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion

Can we make the scroll behaviour more reactive based on the above?

eamansour (commented):
Looking really good overall and the changes work really well locally, just a couple of comments that would be great to address!

galasa-ui/src/components/test-runs/test-run-details/LogTab.tsx
```diff
@@ -162,6 +171,58 @@ export default function LogTab({ logs, initialLine }: LogTabProps) {
...
+      const newRunLog = await fetchRunLog(runId);
+
+      setLogContent(newRunLog);
+
+      // Reset search and filters
+      setSearchTerm('');
+      setDebouncedSearchTerm('');
+      setCurrentMatchIndex(-1);
+      setTotalMatches(0);
+      setSearchCache(new Map());
```
eamansour: If I'm expecting a debug message to appear and I've searched for it in the search bar, then refreshing the run log will clear the searched term. I think it would be good to keep the search bar content and filters the same.

galasa-ui/src/components/test-runs/test-run-details/LogTab.tsx
James-Cocker: Created new const `const ANIMATION_BEHAVIOUR = window.matchMedia('(prefers-reduced-motion: reduce)').matches ? "instant" : "smooth";` making the scroll behaviour match the user's system preference

galasa-ui/src/actions/runsAction.ts
James-Cocker: Great point 😁

eamansour (approved):

PR #264, opened 2026-02-24
James c fix 3270 bugs

### LINKED ISSUE #2552: [BUG] 3270 Refresh Not Working Properly
## Describe the bug
When the 3270 tab is refreshed, it is stuck in an infinite loop due to a race condition from the new tags functionality on the overview page.

## Steps to reproduce
1. Go to 3270 tab
2. Refresh page
3. Notice infinite loading
4. Swap to overview tab
5. Click the edit button
6. Notice the existing tags also have not loaded

James-Cocker:
## Why?

Closes https://github.com/galasa-dev/projectmanagement/issues/2552 by fixed all the mentioned issues.

## Changes
- [x] Moves the tag fetching to the testRunDetails page to avoid the lazy loading from the carbon <Tab> component.
- [x] Makes the max-height of the table of screenshots `!important` to override some odd carbon behaviour (cannot test on localhost but tested this directly in browser)
- [x] Created cache for the api config in `runsAction.ts`, as well as a way to reset it for unit tests
- [x] Fixed 3270 loading issue by ensuring the table is rendered before flicking to the tab (unsure why this works)
- [x] Fixed the URL not flicking to the correct `terminalImage` which was introduced by the above
- [x] Fixed broken unit tests

### DIFF
--- a/galasa-ui/src/actions/runsAction.ts
+++ b/galasa-ui/src/actions/runsAction.ts
@@ -21,6 +21,11 @@ const getCachedApiConfiguration = (): Configuration => {
   return cachedApiConfig;
 };
 
+// Reset the cached API configuration (useful for testing)
+export const resetApiConfigCache = (): void => {
+  cachedApiConfig = null;
+};
+
 export const downloadArtifactFromServer = async (runId: string, artifactUrl: string) => {
   const apiConfig = getCachedApiConfiguration();
   const rasApiClient = new ResultArchiveStoreAPIApi(apiConfig);
--- a/galasa-ui/src/actions/runsAction.ts
+++ b/galasa-ui/src/actions/runsAction.ts
@@ -9,25 +9,9 @@ import { ResultArchiveStoreAPIApi, TagsAPIApi } from '@/generated/galasaapi';
 import { createAuthenticatedApiConfiguration } from '@/utils/api';
 import { fetchRunDetailLogs } from '@/utils/testRuns';
 import { CLIENT_API_VERSION } from '@/utils/constants/common';
-import { Configuration } from '@/generated/galasaapi';
-
-// Cache for API configuration to avoid multiple cookie accesses in the same request
-let cachedApiConfig: Configuration | null = null;
-
-const getCachedApiConfiguration = (): Configuration => {
-  if (!cachedApiConfig) {
-    cachedApiConfig = createAuthenticatedApiConfiguration();
-  }
-  return cachedApiConfig;
-};
-
-// Reset the cached API configuration (useful for testing)
-export const resetApiConfigCache = (): void => {
-  cachedApiConfig = null;
-};
 
 export const downloadArtifactFromServer = async (runId: string, artifactUrl: string) => {
-  const apiConfig = getCachedApiConfiguration();
+  const apiConfig = createAuthenticatedApiConfiguration();
   const rasApiClient = new ResultArchiveStoreAPIApi(apiConfig);
 
   const artifactFile = await rasApiClient.getRasRunArtifactByPath(
@@ -69,7 +53,7 @@ export const downloadArtifactFromServer = async (runId: string, artifactUrl: str
 
 export const updateRunTags = async (runId: string, tags: string[]) => {
   try {
-    const apiConfig = getCachedApiConfiguration();
+    const apiConfig = createAuthenticatedApiConfiguration();
     const rasApiClient = new ResultArchiveStoreAPIApi(apiConfig);
 
     // Note: Tags are already unique from the Set in the frontend, but is checked again by the rest api.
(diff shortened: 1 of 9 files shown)


galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
```diff
@@ -5,15 +5,16 @@
...
 }
 
 .clickableRow {
   cursor: pointer;
 }
 
 .innerScreenshotTable {
-  max-height: min(calc(100vh - 450px), 550px);
+  /* Override Carbon's max-height for the Table component */
+  max-height: min(calc(100vh - 450px), 550px) !important;
```
eamansour: Do we really need to resort to `!important`? There might be another CSS class somewhere that impacts the height of the screenshot table, so tracing through the CSS code might tell us what else would need changing to avoid using `!important`.

It would be good to avoid `!important` since it's a quick and rough way of overriding CSS behaviour.

galasa-ui/src/actions/runsAction.ts
```diff
@@ -9,9 +9,25 @@ import { ResultArchiveStoreAPIApi, TagsAPIApi } from '@/generated/galasaapi';
...
+  if (!cachedApiConfig) {
+    cachedApiConfig = createAuthenticatedApiConfiguration();
+  }
+  return cachedApiConfig;
+};
+
+// Reset the cached API configuration (useful for testing)
+export const resetApiConfigCache = (): void => {
+  cachedApiConfig = null;
+};
```
eamansour: Is this function needed here? It's not used in the production code, so it might be better off moved to the test code if it's only used there?

galasa-ui/src/actions/runsAction.ts
```diff
@@ -9,9 +9,25 @@ import { ResultArchiveStoreAPIApi, TagsAPIApi } from '@/generated/galasaapi';
...
+
+// Cache for API configuration to avoid multiple cookie accesses in the same request
+let cachedApiConfig: Configuration | null = null;
+
+const getCachedApiConfiguration = (): Configuration => {
+  if (!cachedApiConfig) {
+    cachedApiConfig = createAuthenticatedApiConfiguration();
+  }
+  return cachedApiConfig;
+};
```
eamansour: If the API config is cached on the server-side and persists across different requests, could this cause authentication issues between users?

If we use caching, it should be request-scoped and if that's not possible, it should be removed entirely.

James-Cocker: Good point yes, I'll go ahead and remove it as I'm not too sure how i'd go about testing that

galasa-ui/src/actions/runsAction.ts
James-Cocker: Nope, not if we're removing the caching

galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
James-Cocker: Yep used an ID and it all worked fine - not sure why i didn't think of that at the time, thanks!

galasa-ui/src/components/test-runs/test-run-details/TestRunDetails.tsx
```diff
@@ -100,12 +103,14 @@ const TestRunDetails = ({
...
+  };
 
+  useEffect(() => {
     // If 3270 tab has been selected in the URL, move them to the 3270 pannel from the overview page redirection
-    if (is3270TabSelectedInURL && newZos3270TerminalFolderExists) {
+    if (is3270TabSelectedInURL && zos3270TerminalFolderExists && !is3270TabLoading) {
       setSelectedTabIndex(indexOf3270Tab);
     }
-  };
+  }, [is3270TabLoading]);
```
eamansour, line 115: This useEffect is going to complain about `zos3270TerminalFolderExists` and `is3270TabSelectedInURL` not being in the dependency array, should they be added? If not, maybe document this?

galasa-ui/src/styles/test-runs/test-run-details/Tab3270.module.css
```diff
@@ -5,15 +5,16 @@
...
 
 .clickableRow {
   cursor: pointer;
 }
 
-.innerScreenshotTable {
-  max-height: min(calc(100vh - 450px), 550px);
+#innerScreenshotTable {
+  /* Override Carbon's max-height for the Table component */
+  max-height: min(calc(100vh - 450px), 550px) !important;
```
eamansour: `!important` is still here 👀 

We should be able to remove it now :)

James-Cocker: Oops my apologies

eamansour (approved):

PR #269, opened 2026-03-06
Added functionality for header and footer calculation with unit tests

### LINKED ISSUE #2501: Web UI: Scrollability of main content in /test-runs independent to saved queries
## Description
As a Galasa user, i do not want to scroll down the main content of the `/test-runs` page to reach a saved query in a long list of saved queries.

See Slack as an example of this, how an individual chat (on the right) can scroll independently of the list of chats (on the left)

This would solve the flickering of the saved queries box due to the current dynamic resizing implementation.

## Tasks:
- [x] Set the height of the main content to be a fixed proportion of the screen, e.g. `calc(100vh - (height of footer + height of header))` if we wanted the header and footer to always be visible. This solution might not be realistic for smaller screens so will need to be played around with.
- [x] Set height of the saved queries to be the same formula as the height above
- [x] Remove functionality of the saved queries bar to dynamically resize.

James-Cocker:
## Why?

Closes https://github.com/galasa-dev/projectmanagement/issues/2501. 

Sets the max-height for the content on the /test-runs page (including the side bar) to be `100vh-height_of_headers_and_footer` so that they scroll independently of one another, without the observer causing strange flickering and scrolling.

## Screenshot(s) of changes
Notice how the header and footer will always be on the screen.
image
image

## Changes
- [x] Functionality

### DIFF
--- a/galasa-ui/src/components/test-runs/TestRunsDetails.tsx
+++ b/galasa-ui/src/components/test-runs/TestRunsDetails.tsx
@@ -8,6 +8,7 @@ import BreadCrumb from '@/components/common/BreadCrumb';
 import TestRunsTabs from '@/components/test-runs/TestRunsTabs';
 import styles from '@/styles/test-runs/TestRunsPage.module.css';
 import { Suspense, useEffect, useMemo, useRef, useState } from 'react';
+import { getHeightOfHeaderAndFooter } from '@/utils/functions/getHeightOfHeaderAndFooter';
 import useHistoryBreadCrumbs from '@/hooks/useHistoryBreadCrumbs';
 import { useTranslations } from 'next-intl';
 import { NotificationType } from '@/utils/types/common';
@@ -45,8 +46,10 @@ export default function TestRunsDetails({
   const [notification, setNotification] = useState<NotificationType | null>(null);
   const [isEditingName, setIsEditingName] = useState<boolean>(false);
   const [editedName, setEditedName] = useState<string>('');
+  const [maxHeight, setMaxHeight] = useState<string>('68vh');
 
   const inputRef = useRef<HTMLInputElement>(null);
+  const mainContentRef = useRef<HTMLDivElement>(null);
 
   const activeQuery = getQueryByName(queryName);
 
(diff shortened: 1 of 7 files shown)


galasa-ui/src/styles/common/BreadCrumb.module.css
```diff
@@ -5,5 +5,6 @@
  */
 
 .crumbContainer {
+  height: 84px;
```
James-Cocker, line 8: Manually set this height otherwise it seems to default to 64px for an odd reason.

galasa-ui/src/utils/functions/getHeightOfHeaderAndFooter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // Iterate through all elements with this class name
+    for (let i = 0; i < elements.length; i++) {
+      const element = elements[i] as HTMLElement;
+      // Get the offsetHeight which includes padding and border
+      totalHeight += element.offsetHeight;
+      console.log('Hello ' + className + ': ' + element.offsetHeight);
+    }
+  });
+  console.log('Hello ' + totalHeight);
```
eamansour, line 32: Can we remove these log statements if they're not needed?

galasa-ui/src/components/test-runs/TestRunsDetails.tsx
```diff
@@ -58,6 +60,24 @@ export default function TestRunsDetails({
...
+    // Initial calculation
+    updateMaxHeight();
+
+    // Recalculate on window resize
+    window.addEventListener('resize', updateMaxHeight);
+
+    return () => {
+      window.removeEventListener('resize', updateMaxHeight);
+    };
+  }, []);
```
eamansour, line 79: There's some code duplication now with the content of this useEffect and the useEffect in the CollapsibleSideBar component.

Is there a way to reduce this duplication?

eamansour (dismissed):
A couple of minor comments but overall looks good, happy to approve!

galasa-ui/src/utils/functions/getHeightOfHeaderAndFooter.ts
James-Cocker: Oops good spot, thank you!

PR #270, opened 2026-03-09
James c iss2557 remove scroll logic for test runs

### LINKED ISSUE #2557: Web UI: Remove height restrictions on /test-runs
## Story
As a webui user, I want to scroll through the information on `/test-runs` without flickering or odd scrolling issues introduced by the "observer" element, so that the website functions more naturally.

## Background
A 'fix' was initially put in place [here](https://github.com/galasa-dev/webui/commit/2f5ca60b471120e80bf8c9bc2ee379019cff5b6d) using an `observer` element and is currently still in place, causing odd issues occasionally. I started [this](https://github.com/galasa-dev/projectmanagement/issues/2501) story, then realised that people would likely never have as many saved queries as we were thinking they would, which brought on this story.

## Tasks
- [x] Remove previous implementation with observer
- [x] Introduce logic for the bottom few saved queried to invert their menu popup direction so that the page doesn't overflow.

James-Cocker:
## Why?

Closes https://github.com/galasa-dev/projectmanagement/issues/2557. 

Removed height restrictions on elements on the `/test-runs` page as we can assume users will not have excessive saved queries, and the previous implementation was causing odd scrolling issues.

## Changes
- [x] Functionality

### DIFF
--- a/galasa-ui/src/components/test-runs/saved-queries/QueryItem.tsx
+++ b/galasa-ui/src/components/test-runs/saved-queries/QueryItem.tsx
@@ -24,6 +24,7 @@ interface QueryItemProps {
   isCollapsed?: boolean;
   handleEditQueryName?: (queryName: string) => void;
   setNotification?: Dispatch<SetStateAction<NotificationType | null>>;
+  displayMenuUpwards?: boolean;
 }
 
 const ICON_SIZE = 18;
@@ -34,6 +35,7 @@ export default function QueryItem({
   isCollapsed = false,
   handleEditQueryName,
   setNotification,
+  displayMenuUpwards = false,
 }: QueryItemProps) {
   const translations = useTranslations('QueryItem');
   const router = useRouter();
@@ -208,7 +210,7 @@ export default function QueryItem({
         iconDescription={translations('actions')}
         flipped
         className={styles.overflowMenu}
-        align="top"
+        direction={displayMenuUpwards ? 'top' : 'bottom'}
       >
         {actions.map((action) => (
           <OverflowMenuItem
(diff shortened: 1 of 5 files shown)


galasa-ui/src/components/test-runs/saved-queries/QueryItem.tsx
```diff
@@ -46,6 +48,37 @@ export default function QueryItem({
...
+      const wouldOverflow = rect.bottom + MENU_HEIGHT > viewportHeight;
+      setShouldOpenUpwards(wouldOverflow);
+    };
+
+    // Calculate on mount and when the component updates
+    calculateMenuDirection();
+
+    // Recalculate on window resize
+    window.addEventListener('resize', calculateMenuDirection);
+    window.addEventListener('scroll', calculateMenuDirection);
```
eamansour: If we add event listeners for each query item, could this impact the page's performance when someone has a lot of saved queries?

eamansour (approved):

galasa-ui/src/components/test-runs/saved-queries/QueryItem.tsx
James-Cocker: I did give it a go and it seemed okay, and it has a massive advantage over hard coding in the last 3 elements as that will depend on screen size and lots of other bits. Do you have another idea? 😁

eamansour: Perhaps you could look into using the IntersectionObserver or ResizeObserver APIs?

James-Cocker: As mentioned in our slack conversation, we've revised the event listener approach in favour of a one time calculation when the menu is opened. This means that if the use decides to scroll the menu doesn't change direction, however this feels more natural when using the website anyway.

PR #278, opened 2026-03-31
Iss2444 - Fix saved queries not persisting timeframe radio selections

### LINKED ISSUE #2444: Query time options and Save button state not updating correctly when switching between saved queries
## Describe the bug
There are two related issues observed when switching between saved queries with different time configurations.
1. Radio button state inconsistency:
 - Save a query using "Duration 1,0,0" and "Now".
 - Jump to another query configured with "From explicit time" and "To explicit time".
 - Toggle between the two queries using the shortcuts list.
 - The radio buttons for "Duration before To time" and "A specific time" appear to get mixed up or show inconsistent selections.

2. Save button not enabling:
When toggling between "Duration before To time" and "A specific time", the Save Query button does not enable even though the selection has changed.

## Expected behavior
1. Radio button states should reflect the currently active query accurately.
2. The Save Query button should enable when any time configuration is modified.

jadecarino:
## Why?

Fixes https://github.com/galasa-dev/projectmanagement/issues/2444

### Problem
When switching between saved queries with different timeframe configurations, the radio button selections for "From" and "To" time options were not being preserved correctly, causing two problems:

1. Radio button state inconsistency: Switching between queries (e.g., "Duration + Now" to "Specific Time + Specific Time") would show incorrect radio button selections
2. Save button not enabling: The Save Query button would not enable when toggling between radio button options, even though the selection had changed

**Cause of bug:** Radio button selections were stored only in local component state and never persisted to the URL. The component relied on `isRelativeToNow` to infer which radio buttons should be selected, but this approach could not distinguish between all four valid combinations (Duration+Now, Duration+Specific, Specific+Now, Specific+Specific) and did not preserve the saved query configuration when switching between queries.

`isRelativeToNow`:
The `isRelativeToNow` property is still required to control runtime behavior (whether "To" date is dynamic new Date() or fixed) but should have never been used to populate radio button selections when switching between queries. 

## Changes
- [x] Added explicit URL parameters (fromOption and toOption) to persist radio button selections.
   - [x] Added FROM_OPTION and TO_OPTION to TEST_RUNS_QUERY_PARAMS
   - [x] Added fromOption?: string and toOption?: string to TimeFrameValues
- [x] Context now reads fromOption/toOption from URL parameters when loading queries and writes fromOption/toOption to URL parameters when state changes which is persisted with the saved query
- [x] TimeFrameContent now includes radio button selections in state updates and syncs radio buttons from URL parameters when queries load
- [x] Unit test for this scenario has been updated (verifies radio buttons sync correctly when loading queries with explicit parameters) and passing
- [x] Manual testing confirms radio buttons persist correctly across query switches
- [x] Save button enables appropriately when selections change

### DIFF
--- a/galasa-ui/src/components/test-runs/timeframe/TimeFrameContent.tsx
+++ b/galasa-ui/src/components/test-runs/timeframe/TimeFrameContent.tsx
@@ -7,7 +7,7 @@
 
 import styles from '@/styles/test-runs/timeframe/TimeFrameContent.module.css';
 import { TimeFrameValues } from '@/utils/interfaces';
-import { useState, useCallback, useEffect } from 'react';
+import { useState, useCallback, useEffect, useRef } from 'react';
 import TimeFrameFilter from './TimeFrameFilter';
 import { dateTimeLocal2UTC, dateTimeUTC2Local } from '@/utils/timeOperations';
 import { InlineNotification } from '@carbon/react';
@@ -127,6 +127,7 @@ export default function TimeFrameContent({ values, setValues }: TimeFrameContent
   const [selectedToOption, setSelectedToOption] = useState<ToSelectionOptions>(
     ToSelectionOptions.now
   );
+  const isSyncingFromUrl = useRef(false);
 
   const handleValueChange = useCallback(
     (field: keyof TimeFrameValues, value: any) => {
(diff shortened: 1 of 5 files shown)


jadecarino:
Closing as reimplemented in #295

PR #291, opened 2026-04-30
James c iss2029 access token expiry

James-Cocker:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2029. 

This PR supports the upcoming expiry token changes in https://github.com/galasa-dev/galasa/pull/570 for the issue above.

## Screenshot(s) of changes
image

image

image

image

image

## Changes
- [x] When you allocate an access token it asks for a number of days/months/years, or a certain date when it will expire. That expiry time is associated with the user.
- [x] Showing the token in the web UI also shows it's expiry time.
- [x] Any expiry time which is "getting close" will be marked on the UI with a warning.
- [x] Unit tests


galasa-ui/messages/en.json
```diff
@@ -117,11 +119,20 @@
...
+    "custom_lifespan_invalid": "Lifespan must be between 1 and 365 days",
     "error_requesting_token": "Error requesting access token"
   },
   "TokenCard": {
     "createdAt": "Created at:",
-    "owner": "Owner"
+    "expires": "Expires:",
+    "expired": "Expired:",
+    "owner": "Owner",
+    "nearlyExpiredWarning": "Your personal access token will expire soon . Once it expires, you will no longer be able to use it to contact your Galasa service."
```
eamansour: Minor - can the space before the `.` in `... will expire soon .` be removed?

Perhaps also instead of `your Galasa service`, we could change that to say `this Galasa service`?

galasa-ui/src/app/auth/tokens/route.ts
```diff
@@ -15,6 +15,7 @@ export const dynamic = 'force-dynamic';
 
 interface TokenDetails {
   tokenDescription: string;
+  token_lifespan_days: number;
```
eamansour: Minor - can we maintain the camel case formatting here?

galasa-ui/src/app/users/edit/page.tsx
```diff
@@ -5,20 +5,48 @@
...
+      exceededMaximum: true,
+    };
+  }
+
+  return {
+    warningDays: parsedValue,
+    exceededMaximum: false,
+  };
+};
+
```
eamansour: This code is duplicated in the mysettings page. Can we please only have one function that both pages use?

galasa-ui/src/app/users/edit/page.tsx
```diff
@@ -37,6 +65,24 @@ export default async function EditUserPage(props: UsersPageProps) {
...
+      );
+
+      const warningPropertyValue = warningPropertyResponse?.[0]?.data?.value;
+      return getValidatedWarningDays(warningPropertyValue);
+    } catch (error) {
+      return getValidatedWarningDays();
+    }
+  };
+
+  const tokenExpiryWarningConfiguration = await fetchTokenExpiryWarningConfiguration();
```
eamansour: More duplication here - it would be great if we could move these somewhere where both the mysettings and edit user pages can use them

galasa-ui/messages/en.json
```diff
@@ -117,11 +119,20 @@
     "token_name": "Token name",
     "token_name_helper_text": "Use this to distinguish between your tokens in the future.",
     "token_name_placeholder": "e.g. galasactl access for my Windows machine",
+    "token_lifespan": "Token lifespan",
+    "select_lifespan": "Select token lifespan",
+    "custom_lifespan": "Custom",
+    "custom_lifespan_days": "Custom lifespan (days)",
+    "custom_lifespan_helper_text": "Enter a value between 1 and 365 days",
+    "custom_lifespan_invalid": "Lifespan must be between 1 and 365 days",
```
eamansour: Not so sure about having users supply a number for the custom option. Wouldn't it be more user friendly to have them pick a specific date within the next year? Then internally, we could calculate that date from now if we needed a number?

galasa-ui/src/components/tokens/TokenCard.tsx
```diff
@@ -12,39 +12,71 @@ import { AuthToken } from '@/generated/galasaapi';
...
+  const formattedCreationDate = useMemo(() => {
     return formatDate(new Date(token.creationTime!));
   }, [token.creationTime, formatDate]);
+  const formattedExpiryDate = useMemo(() => {
+    return token.expiryTime ? formatDate(new Date(token.expiryTime)) : null;
+  }, [token.expiryTime, formatDate]);
+
+  // Check if token is expired
+  const isExpired = useMemo(() => {
+    if (!token.expiryTime) return false;
```
eamansour, line 37: For security purposes, if a token doesn't have an expiry time, then shouldn't we treat it as expired and force the user to create a new token?

galasa-ui/src/components/tokens/TokenRequestModal.tsx
```diff
@@ -5,24 +5,60 @@
...
   const [error, setError] = useState('');
   const [submitDisabled, setSubmitDisabled] = useState(true);
+  const [selectedLifespan, setSelectedLifespan] = useState<string>(String(PRESET_LIFESPANS[0]));
+  const [customLifespan, setCustomLifespan] = useState<number>(7);
   const tokenNameInputRef = useRef<HTMLInputElement>(undefined);
 
+  const getExpiryDate = (days: number): string => {
+    const date = new Date();
+    date.setDate(date.getDate() + days);
+    return date.toLocaleDateString('en-GB', { day: 'numeric', month: 'long' });
```
eamansour: This is using the en-GB locale, but we should be using the locale that the user has set in their settings.

galasa-ui/src/components/tokens/TokenRequestModal.tsx
```diff
@@ -72,9 +109,12 @@ export default function TokenRequestModal({ isDisabled }: { isDisabled: boolean
         secondaryButtonText={translations('cancel')}
         shouldSubmitOnEnter={true}
         open={open}
+        className={styles.modalContent}
         onRequestClose={() => {
           setOpen(false);
           setError('');
+          setSelectedLifespan(String(PRESET_LIFESPANS[0]));
+          setCustomLifespan(7);
```
eamansour: A magic 7 appears 😄  is this the default lifespan for the "custom" option? Maybe a constant like we've done with the selected lifespan option could go here?

galasa-ui/src/components/tokens/TokenRequestModal.tsx
eamansour: Also, can we have the year displayed in the expiry date as well?

galasa-ui/messages/en.json
eamansour: As an example, this is what the "custom" option looks like in GitHub:

Image

James-Cocker: Yep great point - I was just following along with the mockups

galasa-ui/src/components/tokens/TokenCard.tsx
James-Cocker: I believe that this variable is just used for styling purposes so it shouldn't have any impact on security. I'm thinking if the migration somehow doesn't work, then the tokens probably don't need to all display as expired - but maybe they should as they wouldn't be usable if that were the case.

galasa-ui/src/components/tokens/TokenRequestModal.tsx
James-Cocker: image
How's this?

eamansour: I don't think we need the time in there, just the day, month, and year (with the format adjusted based on user locale)

galasa-ui/src/components/tokens/TokenCard.tsx
eamansour: Yes, I agree that they should all be displayed. That at least gives users a chance to delete old tokens.

eamansour (approved):

PR #295, opened 2026-05-06
Iss2444 - Fix for TimeFrame radio buttons state loading bug

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2444

### Bug

Radio button selections in the Timeframe tab reset to defaults (Duration and Now) when switching between saved queries, instead of reflecting the saved state, as the component TimeFrameContent persisted radio button selections only in local state and never persisted them to the URL.

## Screenshot(s) of changes

https://github.com/user-attachments/assets/759de11b-4984-4747-8353-045ce5d5bdf3

https://github.com/user-attachments/assets/019b0a11-5c3d-48e4-afd0-bbc98f40c9e0

## Changes

- [x] Deleted old component TimeFrameContent and inner component TimeFrameFilter

### Fix

- [x] Added fromSelectionType and toSelectionType to `TimeFrameValues` interface
- [x] Created new `TimeFrameSelector.tsx` component that derives radio button state from URL parameters
- [x] Updated `TestRunsQueryParamsContext.tsx` to encode/decode all 4 radio button combinations
- [x] Added future date prevention on From date picker (`maxDate={new Date()}`)
- [x] Removed obsolete future date validation (better to prevent an error than notify the user after)

### Testing

- [x] New unit tests for URL encoding/decoding and radio button persistence
- [x] All existing tests updated and passing
- [x] Manual testing: Created 4 saved queries of all radio button combinations, verified radio buttons persist correctly when switching queries
- [x] Verified future dates cannot be selected for From time
- [x] Verified URL sharing loads the query's timeframe correctly


galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
```diff
@@ -0,0 +1,398 @@
...
+          ...prevValues,
+          ...finalState,
+          fromSelectionType: fromType,
+          toSelectionType: newToType,
+          isRelativeToNow: newToType === 'now',
+        }));
+      }
+    },
+    [values, fromType, getResolvedTimeZone, translations, setValues]
+  );
```
eamansour, line 258: These handle methods have a lot of overlapping code, it would be great if we could reduce the duplication here.

galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const toType: ToSelectionType = values.isRelativeToNow ? 'now' : 'specificTime';
+  const fromType: FromSelectionType = values.isRelativeToNow ? 'duration' : 'specificTime';
+
+  return { fromType, toType };
+};
+
+/**
+ * Validates timeframe and returns corrected dates with optional notification.
+ */
+export function validateTimeFrame(
```
eamansour: Minor - Is this function used elsewhere? If not, it doesn't need to be exported. 

Also it would be good to maintain consistency when defining functions - above this, we use `const functionName = (...) => {}` but here we're using `export function functionName() {}`

galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+interface TimeFrameSelectorProps {
+  values: TimeFrameValues;
+  setValues: React.Dispatch<React.SetStateAction<TimeFrameValues>>;
+}
+
+/**
+ * Calculates the synchronized state from two valid dates.
+ */
+export const calculateSynchronizedState = (
```
eamansour: I can see this was in the code previously, but it seems strange to have a function defined in a component be imported in a context (TestRunsQueryParamsContext), would it be better to move this to a file in `utils/` and import it from there?

eamansour (commented):
Overall looks good and works well! Just a few suggestions

galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
jadecarino: Sure, I've moved this to /utils/timeOperations.ts as it has other time-related functions, and updated the imports in https://github.com/galasa-dev/webui/pull/295/commits/16eabac0d4a6ede71136a1212bf6860e07237ee1

galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
jadecarino: Thanks, I've removed the `export` and defined this as the other functions above in this file, in https://github.com/galasa-dev/webui/pull/295/commits/16eabac0d4a6ede71136a1212bf6860e07237ee1

galasa-ui/src/components/test-runs/timeframe/TimeFrameSelector.tsx
jadecarino: Thanks Eamonn great spot, I've refactored to remove the duplication in https://github.com/galasa-dev/webui/pull/295/commits/16eabac0d4a6ede71136a1212bf6860e07237ee1

eamansour (approved):

PR #296, opened 2026-05-06
Add side nav for menu navigation on small screens

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2424

## Screenshot(s) of changes
**Before:**
Screenshot 2026-05-06 at 14 00 53

**After:**
https://github.com/user-attachments/assets/7d5fb0e0-10fc-46d3-8c8d-c2e1e716904a

## Changes
- [x] Added sidenav that appears when the webui window becomes too small (handled by Carbon automatically), allowing users to access the "Users" and "Test Runs" pages
- [x] Combined the Galasa logo and "Galasa" header name into a single button rather than having two separate buttons that do the same thing
- [x] Simplified the language selector and theme switcher components to be more consistent while removing unnecessary divs and CSS by using equivalent Carbon components instead of raw HTML elements
- [x] Functionality
- [x] Unit tests

### DIFF
--- a/galasa-ui/src/components/headers/LanguageSelector.tsx
+++ b/galasa-ui/src/components/headers/LanguageSelector.tsx
@@ -6,7 +6,7 @@
 
 'use client';
 
-import React, { useState, useTransition } from 'react';
+import { useState, useTransition } from 'react';
 import { OverflowMenu, OverflowMenuItem } from '@carbon/react';
 import { setUserLocale } from '@/utils/locale';
 import { useLocale, useTranslations } from 'next-intl';
@@ -26,7 +26,7 @@ export default function LanguageSelector() {
     languages.find((lang) => lang.value === locale) || languages[0]
   );
 
-  const [isPending, startTransition] = useTransition();
+  const [, startTransition] = useTransition();
   const router = useRouter();
   const translations = useTranslations('LanguageSelector');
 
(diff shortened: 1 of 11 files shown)


galasa-ui/src/components/headers/PageHeader.tsx
```diff
@@ -44,6 +59,22 @@ export default function PageHeader({ galasaServiceName }: { galasaServiceName: s
...
+          expanded={isSideNavExpanded}
+          isPersistent={false}
+          onSideNavBlur={onClickSideNavExpand}
+        >
+          <SideNavItems>
+            <HeaderSideNavItems>
+              <HeaderMenuItem href="/users">{translations('users')}</HeaderMenuItem>
+              {isFeatureEnabled(FEATURE_FLAGS.TEST_RUNS) && (
+                <HeaderMenuItem href="/test-runs">{translations('testRuns')}</HeaderMenuItem>
+              )}
```
jadecarino: Perhaps we could extract these into a HeaderMenuItems component so there's only one component to update when the list of options grows, and slot that in here and above in the normal HeaderNavigation?

galasa-ui/src/components/headers/LanguageSelector.tsx
```diff
@@ -46,36 +46,34 @@ export default function LanguageSelector() {
...
+      data-floating-menu-container
+      className={styles.overflowMenu}
+      focusTrap={true}
+      align="bottom"
+      flipped
+      renderIcon={() => <Wikis className={styles.renderIcon} />}
+      size="lg"
+      iconDescription={`${translations('tooltip')}: ${selectedLanguage.text}`}
+      aria-label="Filter menu"
+      tooltipAlignment="center"
```
jadecarino: When running the UI in dev mode, there is always a React console error regarding this `tooltipAlignment`: `React does not recognize the `tooltipAlignment` prop on a DOM element.`. As we're in here is there any way we can remove this error?

galasa-ui/src/components/headers/ThemeSelector.tsx
```diff
@@ -26,42 +24,26 @@ const themeOptions: { id: ThemeType; label: string; icon: React.ReactNode; toolt
...
-          >
-            {currentTheme.icon}
-          </button>
-        </Tooltip>
-      </Theme>
-    </div>
+    <HeaderGlobalAction
+      data-floating-menu-container
+      aria-label={next.tooltip}
+      tooltipAlignment="center"
```
jadecarino, line 42: Does this new `tooltipAlignment` introduce a new React error as mentioned in the previous comment? If not, all good!

jadecarino (approved):
Looks great, just a few minor suggestions!

galasa-ui/src/components/headers/PageHeader.tsx
eamansour: Great idea - added a separate component in https://github.com/galasa-dev/webui/pull/296/commits/5d58c0a1b6b0978a46478d89cdc7f89370b262c2

galasa-ui/src/components/headers/LanguageSelector.tsx
eamansour: I'll investigate this now!

eamansour: Turns out, the `tooltipAlignment` property isn't supported in the `OverflowMenu` component (see https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--overview), so I've removed it in https://github.com/galasa-dev/webui/pull/296/commits/6d75c7d0cd80c2be3927426871ea1a165813b336.

galasa-ui/src/components/headers/ThemeSelector.tsx
eamansour: Similar to the above - the `tooltipAlignment` property also isn't supported in the `HeaderGlobalAction` component (see https://react.carbondesignsystem.com/?path=/docs/components-ui-shell-header--overview), so I've removed it in https://github.com/galasa-dev/webui/pull/296/commits/6d75c7d0cd80c2be3927426871ea1a165813b336.

